PR #1, opened 2019-10-07
internal/age: add x25519KeySize constant to replace repeated integer literals

mdlayher:
Signed-off-by: Matt Layher <mdlayher@gmail.com>

A quick tidy up to replace some "magic numbers" that appear in various places in the X25519 code.

### DIFF
--- a/internal/age/x25519.go
+++ b/internal/age/x25519.go
@@ -20,10 +20,13 @@ import (
 	"golang.org/x/crypto/hkdf"
 )
 
-const x25519Label = "age-tool.com X25519"
+const (
+	x25519Label   = "age-tool.com X25519"
+	x25519KeySize = 32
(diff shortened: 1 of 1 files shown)


moritzdietz:
Maybe I have this wrong, but didn't he specifically leave those values to the fixed 32?

FiloSottile:
Yeah, I went back and forth a couple times on this but I really don't like `[x25519KeySize]byte`.

To solve the impasse, I am just switching us to a prototype of golang/go#32670.

Depado:
> Maybe I have this wrong, but didn't he specifically leave those values to the fixed 32?

This is the very definition of a magic number. [See this for reference](https://en.wikipedia.org/wiki/Magic_number_(programming))

PR #5, opened 2019-10-08
.travis.yml: enable Travis-CI

FiloSottile:


.travis.yml
```diff
@@ -0,0 +1,5 @@
+os: linux
+arch: arm64
```
dmitshur: Is using arm64 instead of amd64 intentional?

FiloSottile: Yes, it's irrelevant to the tests, but the new Travis-CI arm64 containers spin up much faster, so there's less wait.

PR #12, opened 2019-10-09
 update readme - added more badges

anatolyyyyyy:
I think it's good idea - add some useful badges


FiloSottile:
Thank you for the PR. I think these are relevant to age developers, not users, so I prefer not having them on the user-facing README.

PR #16, opened 2019-10-25
Add manpage

The-King-of-Toasters:
I showed this off in your last twitch stream, so I might as well make a pull request to make it official. I also remove the reference to `gpg(1)` ;).

Some questions:

 - The manual does not currently contain any copyright preamble. Should I assign copyright to Google like the source files?
- I have omitted the `-aliases` flag as it's not described in the spec. My understanding is that `age -aliases filippo` prints the key for `filippo` and `age -aliases filippo key` modifies the key for `filippo`.
- I haven't documented the `keys.txt` file as I'm unsure of its contents. Will it only contain age keys generated with `-G`, or other key types?

As for the options, I have my own private fork that uses `-G` for key generation, `-D` for decryption and `-A` for alias stuff. I prefer this as it retains the classic UNIX argument style, which the same style used by `signify`/`minisign`. But if you're firm on the existing flags then I don't really mind.

### DIFF
--- a/docs/age.1
+++ b/docs/age.1
@@ -1,0 +1,129 @@
+.Dd October 25, 2019
+.Dt AGE 1
+.Os
+.Sh NAME
+.Nm age
+.Nd encrypt and decrypt data
+.Sh SYNOPSIS
+.Nm
+.Op Fl o Ar file
+.Fl generate
+.Nm
+.Op Fl i Ar file
+.Op Fl o Ar file
+.Ar recipient ...
+.Nm
+.Op Fl i Ar file
+.Op Fl o Ar file
+.Fl d
+.Op Ar identity-file ...
+.Sh DESCRIPTION
+The
+.Nm
+utility encrypts and decrypts data,
+reading from standard input and writing to standard output.
+By default,
+.Nm
+will encrypt input for one-or-more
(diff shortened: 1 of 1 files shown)


docs/age.1
```diff
@@ -0,0 +1,129 @@
...
+.Ev XDG_CONFIG_HOME
+is set,
+else it will look for
+.Pa ~/.config/age/aliases.txt .
+On macOS,
+.Nm
+will look for
+.Pa "~/Library/Application Support/age/aliases.txt" .
+.Sh EXIT STATUS
+.Ex -std
```
xorhash: This should probably be expanded upon since it's not at all clear what "failure" will actually be. In particular, it would be helpful if the exit status differentiated between decryption failure due to a MAC mismatch and everything else (out of memory, etc.) and what to *do* in those crypto failure cases. Though this would need adjusting in the code.

The-King-of-Toasters: Agreed, but I can only work with what I have ;). Let's cross that bridge when we get to it.

The-King-of-Toasters:
This pr is seriously out of date, I'll whip up something new and rebase soon.

Also @FiloSottile pls stream on twitch again :(.

PR #19, opened 2019-11-01
Decrypt private keys encrypted with passphrase.

eliwjones:
This allows one to decrypt age messages encrypted with a `id_rsa.pub` that is associated with a an encrypted `id_rsa` key.

```
$ ./age -d -i /tmp/secret.age ~/.ssh/id_rsa
Enter passphrase for encrypted id_rsa:
```

### DIFF
--- a/internal/age/ssh.go
+++ b/internal/age/ssh.go
@@ -281,14 +281,14 @@ func NewSSHEd25519Identity(key ed25519.PrivateKey) (*SSHEd25519Identity, error) 
 	return i, nil
 }
 
-func ParseSSHIdentity(pemBytes []byte) (Identity, error) {
+func ParseSSHIdentity(name string, pemBytes []byte) (Identity, error) {
 	k, err := ssh.ParseRawPrivateKey(pemBytes)
 	if err != nil {
 		if !strings.Contains(err.Error(), "cannot decode encrypted private keys") {
 			return nil, err
 		}
 
-		fmt.Print("Enter passphrase for encrypted id_rsa: ")
+		fmt.Printf("Enter passphrase for key '%s': ", name)
 		pass, _ := terminal.ReadPassword(int(os.Stdin.Fd()))
 		fmt.Println()
 
--- a/internal/age/ssh.go
+++ b/internal/age/ssh.go
@@ -7,6 +7,7 @@
 package age
 
 import (
+	"bufio"
 	"bytes"
 	"crypto/ed25519"
 	"crypto/rand"
@@ -17,6 +18,8 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"os"
+	"strings"
 
 	"github.com/FiloSottile/age/internal/format"
 	"golang.org/x/crypto/chacha20poly1305"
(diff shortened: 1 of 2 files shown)


magical:
> Sadly, I don't have time today to learn cleanest way to not echo the keyboard input.

Maybe [x/crypto/ssh/terminal.ReadPassword](https://godoc.org/golang.org/x/crypto/ssh/terminal#ReadPassword)?

eliwjones:
> > Sadly, I don't have time today to learn cleanest way to not echo the keyboard input.
> 
> Maybe [x/crypto/ssh/terminal.ReadPassword](https://godoc.org/golang.org/x/crypto/ssh/terminal#ReadPassword)?

Yeah, you're probably right.

I had initial hesitation on it not being necessarily compatible on Windows.. but might as well!

internal/age/ssh.go
```diff
@@ -289,11 +289,8 @@ func ParseSSHIdentity(pemBytes []byte) (Identity, error) {
 		}
 
 		fmt.Print("Enter passphrase for encrypted id_rsa: ")
```
riking: This message should use a user-recognizable key identifier, because you can encrypt non-RSA identities and there may be multiple SSH identities on the system. Can you pass the filename down to here?

internal/age/ssh.go
```diff
@@ -281,7 +284,18 @@ func NewSSHEd25519Identity(key ed25519.PrivateKey) (*SSHEd25519Identity, error)
 func ParseSSHIdentity(pemBytes []byte) (Identity, error) {
 	k, err := ssh.ParseRawPrivateKey(pemBytes)
 	if err != nil {
-		return nil, err
+		if !strings.Contains(err.Error(), "cannot decode encrypted private keys") {
+			return nil, err
+		}
+
+		fmt.Print("Enter passphrase for encrypted id_rsa: ")
```
riking: This message should use a user-recognizable key identifier, because you can encrypt non-RSA identities and there may be multiple SSH identities on the system. Can you pass the filename down to here, or return a detectable error that the caller can use to call a different version of the function?

eliwjones: Since `age` is alpha alpha and the api is in flux, I don't see much value in mucking around with the internals.

Were I designing the api, I'd make it completely explicit.  And usage for rsa encrypt decrypt would not do any auto-detecting.

It would just be usage like:
```
echo "secret messages" | ./age "`curl -s https://github.com/eliwjones.keys`" > /tmp/secret.age
```
and:
```
./age -d -i /tmp/secret.age ~/.ssh/id_rsa
```

I guess one could make it more complex and have the encrypted file have information in the header on what public key to use for decryption.. if FiloSottile has plans for that, I'd imagine he will have opinions on how exactly he wants to handle encrypted `id_rsa` files.


This PR is really just for anyone whose mucking about and was wondering how to use their encrypted `id_rsa` with it.

eliwjones: Ha, as is my wont.. I say "No", and then immediately make the change.

Was trivial to pass down the `name` argument and be explicit on what we are asking for the password for.

magical:
Looks like this was implemented in c624abc0ad4f15d737d3bfd74de970227647d2f0.

FiloSottile:
Yeah, sorry about that. As I mentioned in the README, everything is still a little too fluid to take PRs.

eliwjones:
> Yeah, sorry about that. As I mentioned in the README, everything is still a little too fluid to take PRs.

Ha, no hard feelings @FiloSottile 

I did it mainly for myself to be able to experiment with my password protected key and figured someone might find it useful in the meantime.

I figured you'd have a more exacting approach that would work better with the style, structure of Age.

Keep going!

Being able to send encrypted messages to people based off of their public github keys is going to have a much larger impact that I think many people realize.

PR #20, opened 2019-11-25
support for parsing encryption key from file

sigmarelax:
Tested with multiple file arguments and with multi-lined files. In compliance with the published spec document.


FiloSottile:

PR #28, opened 2019-12-28
Provide a `--output` option.

noncombatant:
For consistency across all options.


FiloSottile:
Thanks! I admit my eye twitched when I wrote that usage.

PR #31, opened 2019-12-28
Update the installation and build instructions.

noncombatant:


FiloSottile:
`go build filippo.io/age/cmd/...` will build both binaries, but there was a typo in it, thank you!

Issue #25, opened 2019-12-27
Consider goreleaser for building releases


twpayne:
[goreleaser](https://goreleaser.com/) makes it really easy to build Go binaries for multiple platforms, including generating `.rpm`s, `.deb`s, tarballs, zip files, Homebrew formulae, Snapcrafts, Scoops, and the like. It would be a great fit for age.

Let me know if you'd like a PR that adds this. Thanks to @joemiller who contributed this to chezmoi in https://github.com/twpayne/chezmoi/pull/33 - I'd like to pay the contribution forward.

emaste:
It looks like there's not yet FreeBSD support in [goreleaser](https://goreleaser.com) :(

twpayne:
@emaste Goreleaser does support FreeBSD, and OpenBSD. It supports all OSs that Go can cross-compile to.

whalehub:
@FiloSottile Any thoughts on this?

PR #50, opened 2019-12-29
all: add initial goreleaser config

twpayne:
Initial step towards #25, ready for initial review.


FiloSottile:
Thank you! I went with a simpler GitHub Action for now, but this reminds me that I need to include the LICENSE in the tarball.

PR #43, opened 2019-12-28
Get recipient keys from https:// and file:// URLs

paulc:
This patch includes support for https:// and file:// URLs as recipient arguments. The URL is fetched and each line is added as a recipient.

e.g.

    -r https://github.com/<user>.keys       (Use GitHub keys)
    -r file:///home/<user>/.ssh/<key>.pub   (Use local SSH key)

This is mostly useful to allow recipient keys to be specified directly from a GitHub URL (or other service which provides an equivalent)

PR #64, opened 2019-12-31
Add support for GitHub recipient

saltsa:
This PR adds support to fetch user public keys directly from Github API.  (edit: seems to be duplicate of #43)


Mohammed90:
Doesn't #43 supercede this by being more generic?

saltsa:
> Doesn't #43 supercede this by being more generic?

Yes, that's true. That PR likely needs support for "github:username" syntax, otherwise it supercedes. But this PR can be closed if that's going to be used!

FiloSottile:

PR #76, opened 2020-01-02
README: add Arch Linux AUR instructions.

mfs:
Update readme with Arch Linux AUR instructions.


README.md
```diff
@@ -98,4 +98,12 @@ On OpenBSD -current and 6.7+, you can use the port:
 pkg_add age
 ```
 
+On Arch Linux there are [AUR](https://aur.archlinux.org) packages for age and age-git.
+```
+auracle clone age       # or whichever AUR helper you use
```
alerque: If you are going to suggest the full manual build method why suggest an AUR helper here at all? Just show a regular git clone. Those that want to use helpers often use ones that to all the other steps too. For example `yay -S age` ... don.

Tharre:
Adding any instructions at all seems bad, either the person is already familiar with the AUR, in which case they're redundant. Or they have never used the AUR, in which case teaching people to blindly trust AUR packages is a bad idea.

alerque:
Maybe just linking the respective AUR package pages would be an acceptable compromise.

FiloSottile:
Took the compromise from the mkcert README, of giving a link and the manual installation instructions, so we don't have to pick a helper.

PR #80, opened 2020-01-03
Forgot decryption flag in example

fubuloubu:


FiloSottile:
Ah sorry I merged #85 before I saw this.

Issue #57, opened 2019-12-30
Output files are created even if encryption/decryption failed


wgslr:
## Environment

* OS: Arch Linux 5.4.6-arch3-1 #1 SMP PREEMPT Tue, 24 Dec 2019 04:36:53 +0000 x86_64
* age version*: 1.0.0beta2-1

## What were you trying to do
Decrypt a file without providing the `-i` argument and subsequently fix the mistake.

## What happened
The first decryption attempt already created an empty output file, and a next invocation with fixed argument refused to overwrite it.

```bash
$ echo 'Hello age' | age -a -r 'age1fh296r26vut9rpdarl89765z5qh9pavgmlam3w9s69uh5t82darqat8yat' -o encrypted.age.ascii
$ age -d -o decrypted.txt encrypted.age.ascii
Error: no identity matched a recipient
[ Did age not do what you expected? Could an error be more useful? Tell us: https://filippo.io/age/report ]
# fix the mistake
$ age -d -o decrypted.txt -i key.txt encrypted.age.ascii 
Error: failed to open output file "decrypted.txt": open decrypted.txt: file exists
[ Did age not do what you expected? Could an error be more useful? Tell us: https://filippo.io/age/report ]
$ ls -lh
total 8.0K
-rw-r--r-- 1 wojciech wojciech   0 Dec 30 01:52 decrypted.txt
```
### Encryption
Creation of empty files is observed also when encrypting, when given incorrect `-r` argument:

```bash
$ echo 'Hello age' | age -a -o encrypted -r BAD
Error: unknown recipient type: "BAD"
$ ls -l encrypted
-rw-r--r-- 1 wojciech wojciech 0 Dec 30 02:02 encrypted
```

## UX side
Apart from the incorrect, I assume, behaviour of creating empty files, I am not a fan of the no-overwrite policy. I think it would be better to match most unix tools' behaviour of overwriting by default or add an `-f/--force` flag.

\* would be nice to have `age --version` to properly check that

Issue #22, opened 2019-11-27
CLI user interface?


fd0:
Hi,

thanks a lot for writing age, I'm really looking forward to using it!

Is the CLI user interface already fixed, or are you open to a discussion about it?

I'd like to propose the following:

 * Add subcommands (like `git` and `go`):
   * `age encrypt` (maybe default?)
   * `age decrypt`
   * `age generate`
 * Use the double-dash syntax (maybe with `pflag`?) to distinguish between short and long options

With these proposed modifications, the commands from the spec would look as follows:

    $ age generate > key.txt

    $ echo "_o/" | age encrypt pubkey:98W5ph53zfPGOzEOH-fMojQ4jUY7VLEmtmozREqnw4I > hello.txt.age

    $ age decrypt key.txt < hello.txt.age
    _o/

    $ tar cv ~/xxx | age encrypt github:Benjojo github:FiloSottile | nc 192.0.2.0 1234

    $ echo "_o/" | age encrypt -o hello.age pubkey:98W5ph53zfPGOzEOH-fMojQ4jUY7VLEmtmozREqnw4I

    $ age encrypt -i hello.txt -o hello.txt.age -p
    Type passphrase:

The upsides are:
 * Adding subcommands would avoid ambiguities by forcing the user to be explicit about the intended operation. With `gpg`, I'm never entirely sure if it is in encrypt or decrypt or signing or whatever mode. In addition, this makes it easy to distinguish between a command (`generate`, `encrypt`) and an option (`--output`, `--recipient` etc.).
 * Make it clear to the user that subcommands are mutually exclusive: `age encrypt generate -o foo` is clearly wrong, but `age -encrypt -generate -o foo` is not so clear in my opinion.

For the flag parsing, I really like http://github.com/spf13/pflag, we could also implement that ourselves. Subcommands can be implemented without an external dependency. What's your take on dependencies for `age`? Stdlib only?

What do you think? I'm willing to help set this up if you're up for it, if not please just close this issue and carry on. Thanks!

FiloSottile:
About long options, I always planned to provide long versions of all short options except `-i` and `-o`. The standard flags package already supports double dashes if that's what you prefer, so I don't see the need to use a third party package: the only difference would be that it would make it possible to combine multiple short options, which I think is an anti-pattern. We shouldn't have that many options to begin with!

[...]

PR #97, opened 2020-01-28
Add --version and --help flags to executables

### LINKED ISSUE #74: UX: Show usage on naked command
## What were you trying to do

Show the usage documentation.

## What happened

```
$ age
Error: missing recipients.
Did you forget to specify -r/--recipient or -p/--passphrase?
```

## Additional detail

Right after installing age, most people are going want to see the usage output. It makes for a better user experience if you show the usage / help for all things that the user might try, instead of making the user try multiple different flags to get the usage to appear.

I'd suggest displaying the usage output when age is run with any of the following arguments:

- `age`
- `age help`
- `age --help`
- `age -h`

### LINKED ISSUE #101: UX: What version of age am I running?
How can I know what version of age I'm using? There doesn't seem to be any `-version` flag.

RKinsey:
Closes #74 , Closes #101 
This patch adds flags for access to the usage message and the version tag/hash as requested in #74 and #57, respectively.  

`age` and `age-keygen` with the `--help` or `-h` flag will print their usage message and exit, and `age` will also print usage when no flags or arguments are passed. Per #22, I didn't add `age help`.    
`age-keygen` now has a usage message.  

~~`-v` and~~ (Removed, see below) `--version` work for both `age` and `age-keygen`, and will print the most recent git info at time of compilation. There are two build scripts (bash and Powershell) that grab the most recent tag and current short hash, and uses -ldflags to pass them to main.version and main.commit (named for compatibility with goreleaser defaults). Current format is  
``` 
Version: [most recent tag]
Hash: [shorthash] 
```   
and  
`Version: [most recent tag]` when on a tagged commit or the --release flag is passed to the script.  
I also updated the Homebrew formula for this, but it doesn't currently have an effect


wgslr:
I'm glad to see you implemented my proposal. Just one note: I would be wary of using the `-v` flag for `--version`, as it is often used for `--verbose` and I could image `age` adding such flag in the future. I suggest either using `-V` or not providing a short flag at all, as the version invocation is done rarely and does not need to be particularly compact.

RKinsey:
Good point, removed `-v`

alerque:
@RKinsey I think you can also add `Closes #101` to this issue description.

RKinsey:
Got it

Fastidious:
Would be great if this PR could be accepted. What say you, @FiloSottile? :-)

brad2014:
Just a thought: There are two relevant version strings related to the `age` app: the release version of the code (currently "v1.0.0-beta2") and the version(s) of the supported file spec (currently "v1", but a later release might support multiple file specs).

Do you want to consider wording the `age --help` and the `age --version` output to distinguish between these two uses of the word version?  Maybe:

```
% age --version 
Release version: 1.0.0-beta2
Age spec version: v1
```

shepting:
I'd love to see this land! Any updates @RKinsey

FiloSottile:
Most users will eventually use `go install filippo.io/age/cmd/...@latest` (which from Go 1.16 embeds the version) or Homebrew or a package manager, so I took a slightly simpler path in #173. Thank you!

@brad2014 that's a good point, but we don't plan to ever drop support for older versions of the spec, so for now I think it's fine not to expose it.

ank-everstake:
I think it may be better to expose info for debug purposes.
User can open age-encrypted file on another machine with different age-binary. 
So basically if we hide it we don't have any option to know what binary the user has. 
You can compile by yourself, get pre-built binaries or get them through pacman AUR (in Arch).
Please consider to add the version back. Thanks.

CC @dd-everstake

PR #103, opened 2020-02-21
Add pronunciation tip to the README

jsha:
Since some folks may use the tool without clicking through
on the spec and learning about pronunciation there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # age
 
-age is a simple, modern and secure file encryption tool.
+age (pronounced ah-geh, IPA [age] - the e is not silent) is a simple, modern and secure file encryption tool.
 
 It features small explicit keys, no config options, and UNIX-style composability.
 


alerque:
Ummm...this doesn't seem right. This PR was a good suggestion, but the commit that marked it as closed is a terrible replacement. No offense but an offsite link to a Google Translate page from Italian to "my language" (turns up Turkish for me) is completely not useful. As an end user I wouldn't appreciate that at all, and honestly would be totally confused. It would seem like a mistake. I suggest re-evaluating the initial contribution here and including the explanation inline. You can still something link for the audio (I assume that was the goal?) but it should be explained in so many words first or the link will just muddy the waters.

waldyrious:
The description in the [google doc](https://docs.google.com/document/d/11yHom20CrsuX8KQJXBBw04s80Unjv8zCg_A7sPAX_9Y/preview) seems perfectly suitable to the README IMO:

> It’s called “age”, which might be an acronym for Actually Good Encryption, and it’s pronounced like the Japanese [上げ](https://translate.google.com/?sl=ja&tl=en&text=%E4%B8%8A%E3%81%92&op=translate) (with a hard g).

(And the Google Translate link used in it makes more sense in context.)

FiloSottile:
Hmm, I didn't merge the original because I don't want the pronunciation to be half of the first line in the README. It's not that important. Gave it another try.

Issue #15, opened 2019-10-11
Default keys and aliases location on non-Unix platforms


str4d:
The specification gives the Unix path `~/.config/age/*` as the default location it looks for keys and aliases. How should this map to Windows and macOS? The [directories](https://crates.io/crates/directories) Rust crate (which aims to follow per-OS conventions) appears to indicate the following as being approximately OS-shaped:

- Windows: `C:\Users\Alice\AppData\Roaming\age tool\age\config\*`
- macOS: `/Users/Alice/Library/Preferences/com.age-tool.age/*`

We could also go simpler by just using the known folders for program settings as the base:

- Windows: `C:\Users\Alice\AppData\Roaming\age\*`
- macOS: `/Users/Alice/Library/Preferences/age/*`

Thoughts? Is there somewhere else these should go? Are users expected to be editing these by hand (in which case maybe a Documents-like folder should be used, which is more accessible)?

danlamanna:
FWIW, It looks like the Go team had a lot of [discussion](https://github.com/golang/go/issues/29960) on this topic for the 1.13 release and produced [os.UserConfigDir](https://golang.org/pkg/os/#UserConfigDir). Perhaps `age` should rely on that?

str4d:
I scanned the discussion there and it seems like a reasonable approach. For my Rust impl to be compatible I'd need to manually detect macOS and handle it separately because of how the `directories` crate differs, but that's not too bad.

Fastidious:
I think `age` should use `$home/.config/age/` under macOS as well. Several applications share the same directory.

PR #104, opened 2020-02-22
separate public/private key files

### LINKED ISSUE #91: UX: separate public/private key files
## What were you trying to do

Sharing the public key.

## What happened

I needed to open the file and copy-paste the public key characters in order to share the public key. Having a separate file for the public key would make it more easy for me to share using standard tools, avoids copy-pasting user errors and avoids shoulder surfing issues.

downneck:
Closes #91 
"age-keygen" with no flags will generate ~/.age/me.pub and ~/.age/me.key
"age-keygen -o filename" now generates ~/.age/filename.pub and ~/.age/filename.key
"age -d" will load ~/.age/*.key as well as ~/.ssh/id_rsa and ~/.ssh/id_ed25519
"age -r me" will encrypt using the key found in ~/.age/me.pub
"age -r marysmith" will encrypt using the key found in ~/.age/marysmith.pub


str4d:
You should use `~/.config/age` to match the spec regarding age user data storage. See also #15.

downneck:
@str4d thanks! updated.

alxiong:
I'm wondering whether all pub keys should be put into a single file with annotated alias or should they be in separate files

@FiloSottile already had a proposal in his doc:
> Encryption to an alias (stored at ~/.config/age/aliases.txt, change with -aliases)
$ cat ~/.config/age/aliases.txt
filippo: pubkey:jqmfMHBjlb7HoIjjTsCQ9NHIk_q53Uy_ZxmXBhdIpx4
ben: pubkey:ZAE2ZnRdItykp0ncAZJ2FAzIIfTvmGcgIx/759QhnQw github:Benjojo
$ tar cv ~/xxx | age -r alias:filippo > xxx.tar.age

I feel that combing the _alias_ approach is more concise and easier for #110 (if applicable). 

If so, then `age-keygen -a myalias` would _append_ an entry in `~/.config/age/aliases.txt` and putting private key in `~/.config/age/myalias.key`

FiloSottile:
We went through a few different ideas for how to add these conveniences, but ultimately none of them felt right because age is meant to be an explicit tool without opinions on how keys are exchanged or managed.

We ended up removing aliases, as well as default identities, and we don't have a `.config` path anymore. Applications and users should manage keys as they prefer.

angelacastaneda:
for anyone missing the conveniences of default keys, I made a [shell script](https://gist.github.com/angelacastaneda/c700019600d745698acc028f9f8bac97) that tries to emulate the gpg experience.

i.e.

* `age -e <file>` produce an ascii armored <file>.age file in the same directory encrypted with the `$age_def_rec` file
* `age -d <file>` decrypts the file to stdout with the `$age_def_id` file
* anything that's not a file or doesn't uses -e/-d flags gets sent to the proper age binary location at `$age_bin`

PR #112, opened 2020-03-26
Added debian control files for packaging in Debian/Ubuntu/etc.

pauldotknopf:
I'm not sure if you'd even want them, but I see no harm in trying to put them upstream :)


alerque:
I think there is some harm in having downstream packaging files in upstream project repos. Expect in the case where the project owner is also the primary responsible party for downstream packaging, they are almost never in sync. Basically downstream(s) need to keep that info in a repository of their own anyway and the upstream stuff is just a template to get started that may or may not be followed exactly. In the case of Debian for example the changelog _must_ be edited for every rebuild which includes tweaking what distro targets it builds for, rebuilding for reasons unrelated to the upstream, etc. Inevitably this ends up being stale.

Also I don't think you'd want the clutter of including packaging templates for all distros! A PKGBUILD for Arch and a Ruby formula for Homebrew and a ...

pauldotknopf:
Yeah, you are right. If nothing else, it's now known that it exists in this fork if anyone needs it.

PR #114, opened 2020-04-10
export the minimum from internal/age

tgulacsi:
Is it possible to use age as a library?
To decrypt files encrypted with "age" tool and encrypt such it can decrypt?
I don't want to embed the "age" executable...


dncohen:
FYI, I posted #119 to accomplish the same thing.  I posted it before I saw this.  But also #119 is a smaller change that doesn't change the domain of the package.

PR #115, opened 2020-04-20
Support remote URLs and "github://", "file://" schemes to fetch recipient public keys

### LINKED ISSUE #84: RFE: Allow use of ssh public key *path* rather than key contents for recipient
## Environment

* OS: fedora 30
* age version: e43cf8b4a2d571df8dcb30783a460acbf188adb4

## What were you trying to do

I wish that I could encrypt to a public key by specifying a path to the public key file:

    age -r ~/keys/someperson.pub

Rather than the *contents * of the public key file:

    age -r $(cat ~/keys/someperson.pub)

(Maybe this would require a new option (`-R`?) rather than overloading the behavior of the existing `--recipient/-r` option)

And obviously it would be nice if this same feature were available for `age` public keys as well.

aerth:
This PR supersedes #43 and #64, adding support for `github://`, `file://` and URL schemes (`https://`) for the `-r` (RECIPIENT) flag

Closes #84 

Example usage:

```
echo "hello" | age -r github://aerth -o /tmp/encrypted.txt
echo "hello" | age -r https://github.com/aerth.keys -o /tmp/encrypted.txt2
echo "hello" | age -r file://$HOME/.ssh/id_ed25519.pub -o /tmp/encrypted.txt3
echo "hello" | age -r "$(cat $HOME/.ssh/id_ed25519.pub)" -o /tmp/encrypted.txt4
```

Thank you @paulc for laying the framework for the `github://` scheme


aerth:
worth noting that this introduces _networking_ and uses default `net/http` client, direct connection (no proxy, system resolver)

someone could later add proxy support via environmental variable or flag

nxg:
A problem here is that the URI prefix will surely frustrate tab-completion.  That is, I can type `age -r ~/.ssh/id_r^I` and have the recipient filename complete, but I can't type `age -r file://~/.ssh/id_r^I` and expect it to complete.  I _can_ do `age -r file://$HOME/.ssh/i^I` and it completes (somewhat to my surprise; this works in bash, but not in eg zsh), but that's more typing.

It's a minor usability issue, certainly, but it's enough of an issue that it makes me feel as if I'm missing something.

sylr:
https://github.com/hashicorp/go-getter might be relevent to this.

FiloSottile:

Issue #63, opened 2019-12-31
Expose Go library functions (please review the API!)


schollz:
## What were you trying to do

Use age encryption in a Go program by importing it.

## What happened

Because `age` uses `internal` it [prevents importing](https://golang.org/doc/go1.4#internalpackages). I suspect this was on purpose, and that's okay, but just wanted to check if this was sought a possibility for this spec.

ovalseven8:
According to the **[`age` documentation](https://docs.google.com/document/d/11yHom20CrsuX8KQJXBBw04s80Unjv8zCg_A7sPAX_9Y/preview)** it's intended to be a library:
> This is a design for a simple file encryption CLI tool, Go library, and format.

xt0fer:
perhaps we're just getting caught by the 1.4/1.5 design stricture? If `internal` was renamed to `library` or `impl` would that change the intention of the project?

schollz:
@ovalseven8 That's great! I should have been more specific - I'm actually wondering if this implementation (`filippo.io/age`) is meant to serve as the library.

schollz:
@xt0fer exactly, any name change would allow it to be used as a library. However it was probably purposely done to be `internal`, just wanted to double check

abarisani:
I'd welcome this for attempting to compile this under TamaGo (https://github.com/inversepath/tamago).

schollz:
I made this for now: https://github.com/schollz/age/ but I'd rather not have to merge upstream to keep it up to date, so it would be nice if this repo is the library.

@FiloSottile, do you plan on having this repo be used as a library?

sashabaranov:
Exposing age as a library would very much benefit building mobile apps with good encryption through gomobile!

richo:
Is there any word on this? I'd love to avoid shelling out to age if I can avoid it.

cyb3rz3us:
Apologies in advance for not fully understanding the ask here but since 'age' is using the Go crypto library, why would someone then want to have 'age' be a crypto library for use in a  Go program?

I guess I would think one would just use the same libs that 'age' uses for their Go program...No?

hairyhenderson:
@cyb3rz3us it's not "just" using the Go crypto package... it's also implementing [a specific format](https://age-encryption.org/v1).

The simplest (and probably most naïve) solution to this issue would be to expose all the packages that are currently in https://github.com/FiloSottile/age/tree/master/internal; but I don't think @FiloSottile _really_ wants to do that 😉...

[...]

PR #119, opened 2020-05-01
expose age encryption library

dncohen:
These changes expose age/internal/age, making an importable package that other projects can use.  As requested in issue #63 

Perhaps there are reasons not to expose this in its current state.  But I have a project that could use a library like this, so, my fingers are crossed that this PR is well received!


FiloSottile:
Thank you for contributing! I wanted to make some changes to the API before exposing it. I am mostly done, so could you check that the current age, agessh, and armor packages would work for you, and comment on #63?

PR #120, opened 2020-05-01
Export the minimum from internal/age

### LINKED ISSUE #63: Expose Go library functions (please review the API!)

tgulacsi:
Fixes #63 

Simpler than #119 :-P - exposes only the bare minimum, and leaves everything under internal/age.
Does not change the domains as #114.

### DIFF
--- a/age.go
+++ b/age.go
@@ -1,0 +1,66 @@
+// Copyright 2019 Google LLC
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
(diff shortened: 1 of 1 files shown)


FiloSottile:
Thank you for contributing! I wanted to make some changes to the API before exposing it. I am mostly done, so could you check that the current age, agessh, and armor packages would work for you, and comment on #63?

PR #131, opened 2020-06-28
Manual pages

jamesog:
Inspired by https://twitter.com/FiloSottile/status/1277084905428656129 I thought I'd quickly throw together a couple of man pages, based on the README. :-)


cmd/age-keygen/age-keygen.1
```diff
@@ -0,0 +1,25 @@
...
+.Op Fl o Ar file
+.Sh DESCRIPTION
+.Nm
+generates an
+.Nm age
+secret key.
+.Pp
+If the
+.Fl o
+flag is not specified the key is output to standard output with the public key
```
xorhash, line 17: Grammar nit: a comma after “specified” is required here due to the prefixed “if” clause as far as I know.

cmd/age-keygen/age-keygen.1
```diff
@@ -0,0 +1,25 @@
...
+as a comment.
```
xorhash, line 18: This leaves out what the happens with **-o**: What is stored where?

Experienced users will know that **-o** normally means “output file”, sure, but perhaps it'd be useful to specify what it does for all those who aren't familiar with this convention.

Maybe I'm being too pedantic here though.

cmd/age-keygen/age-keygen.1
```diff
@@ -0,0 +1,25 @@
+.Dd June 28, 2020
+.Dt AGE-KEYGEN 1
+.Sh NAME
```
xorhash, line 3: This may be invalid mdoc.

According to [mdoc(7)](https://man.openbsd.org/mdoc.7#Os~2), `.Os` is the mandatory third macro for any mdoc file. groff_mdoc(7) is less clear about this; the “TITLE MACROS” section only seems to imply it as convention, rather than necessity.

Also affects `age.1`.

cmd/age/age.1
```diff
@@ -0,0 +1,162 @@
...
+.Op Fl a
+.Op Fl o Ar output
+.Op Ar input
+.Nm
+.Fl -decrypt
+.Op Fl i Ar key
+.Op Fl o Ar output
+.Op Ar input
+.Sh DESCRIPTION
+In the first synopsis form,
```
xorhash, line 18: Nit: Wouldn't “If **--decrypt** is not given” be easier to read and understand? That would avoid people having to refer back to the synopsis section.

Similarly for line 24.

cmd/age/age.1
```diff
@@ -0,0 +1,162 @@
...
+In the second synopsis form,
+.Nm
+decrypts
+.Ar input .
+.Pp
+The following options are available:
+.Bl -tag -width Ds -compact
+.It Fl o , Fl -output Ar output
+Write the result to the file at path
+.Ar output .
```
xorhash, line 33: Something that people might want to know: Does this *overwrite* output or error out if the file already exists?

Just food for thought.

cmd/age/age.1
```diff
@@ -0,0 +1,162 @@
...
+.It Fl a , Fl -armor
+Encrypt to a PEM encoded format.
+.It Fl p , Fl -passphrase
+Encrypt with a passphrase.
+.It Fl r , Fl -recipient Ar recipient
+Encrypt to the specified
+.Ar recipient .
+Can be repeated.
```
xorhash, line 41: Does this mean I can do **-r** multiple times for the *same* key (just without getting an error about it) or multiple times to encrypt to *multiple recipients*?

Similar thoughts for **-i**.

This is described later and could probably be omitted here entirely because of that.

cmd/age/age.1
```diff
@@ -0,0 +1,162 @@
...
+Decrypt the input to the output.
+.It Fl i , Fl -identity Ar key
+Use the private key file at path
+.Ar key .
+Can be repeated.
+.El
+.Pp
+The
+.Ar input
+argument is optional and defaults to standard input. Only a single file may be
```
xorhash, line 52: Missed newline after period. This is necessary for sentence spacing, though I believe you're likely already aware of this and just missed it.

cmd/age/age.1
```diff
@@ -0,0 +1,162 @@
...
+The identity
+.Ar key
+supplied to
+.Fl i
+/
+.Fl -identity
+is a path to a file with
+.Nm
+secret keys, one per line
+(ignoring "#" prefixed comments and empty lines), or to an SSH key file.
```
xorhash, line 84: Did you mean to use `.Qq` here or `.Sq` since it's a single characters instead of ""?

cmd/age/age.1
```diff
@@ -0,0 +1,162 @@
...
+Multiple keys can be provided, and any unused ones will be ignored.
+.Ss Multiple recipients
+Files can be encrypted to multiple recipients by repeating
+.Fl r
+/
+.Fl -recipient .
+Every recipient will be able to decrypt the file.
+.Ss Passphrases
+Files can be encrypted by using
```
xorhash, line 93: “encrypted with a passphrase”? Without the heading, this sentence seems to imply that any other options don't encrypt files.

cmd/age/age.1
```diff
@@ -0,0 +1,162 @@
...
+.Fl p
+/
+.Fl -passphrase .
+By default
+.Nm
+will automatically generate a secure passphrase. Passphrase protected files are
```
xorhash, line 99: Missed newline after period.

cmd/age/age.1
```diff
@@ -0,0 +1,162 @@
...
+.Po
+.Nm ssh-agent
+is not supported.
+.Pc
+.Pp
+Note that SSH key support employs more complex cryptography, and embeds a
+public key tag in the encrypted file, making it possible to track files that
+are encrypted to a specific public key.
+.Sh EXAMPLES
+Generate a secret key, encrypt data, and decrypt:
```
xorhash, line 118: I'm not sure if this was secret brilliance on your part or just a coincidence of copying over the README, but this conveniently also encourages people just following the examples blindly to frequently generate new keys, which seems like a net win to me.

cmd/age/age.1
```diff
@@ -0,0 +1,162 @@
...
+$ age-keygen -o key.txt
+Public key: age1ql3z7hjy54pw3hyww5ayyfg7zqgvc7w3j2elw8zmrj2kg5sfn9aqmcac8p
+$ tar cvz ~/data | age -r age1ql3z7hjy54pw3hyww5ayyfg7zqgvc7w3j2elw8zmrj2kg5sfn9aqmcac8p > data.tar.gz.age
+$ age -d -i key.txt -o data.tar.gz data.tar.gz.age
+.Ed
+.Pp
+Encrypt
+.Pa example.jpg
+to multiple recipients and output to
+.Pa example.jpg.age:
```
xorhash, line 130: Did you mean to add a space in front of the colon?

jamesog:
Thanks @xorhash, great feedback! I've updated based on your suggestions. Let me know how it looks now.

cmd/age/age.1
jamesog: I shall attribute the brilliance here to Filippo. :-)

xorhash:
Looks shippable to me now. Thanks for sorting this out!

(Compare and contrast with #16 if you're bored.)

benjojo:
This is awesome, thanks James

johanfleury:
FYI, while preparing a Debian package for age, I created man pages from the `--help` output, you can take inspiration from those if needed: <https://salsa.debian.org/Arcaik/age/-/tree/debian/sid/debian/man>.

Fastidious:
@johanfleury isn't **AUTHOR** information normally added at the end of man pages?

IcedQuinn:
is there anything left that needs to be done for this?

Fastidious:
How does this works on platforms other the *nix?

IcedQuinn:
> How does this works on platforms other than *nix?
Mandoc has options to output HTML files. One would use that for non-Unix distributions.

FiloSottile:
Thanks everyone! I finally expanded the man page and merged it!

PR #140, opened 2020-09-04
increase decryption recipient limit from 20 to 1000

### LINKED ISSUE #139: Reconsider 20 recipients limit
## What were you trying to do

Trying to decrypt a file with 69 recipients.

## What happened

```
$ age --decrypt --identity /home/ryantm/.ssh/id_rsa monit.pem
Error: too many recipients
```

## What I expected

The file gets decrypted.

I found a part of the code where it limits the recipient number of decrypted messages to 20: https://github.com/FiloSottile/age/blob/0c650f815dde10d6e1644219b26615c5742ef743/age.go#L127

Can we expect that `age` will be limited to 20 recipients or is this a temporary limitation?

ryantm:
closes #139 

I tested decrypting my file with 69 recipients and it worked.

### DIFF
--- a/age.go
+++ b/age.go
@@ -124,8 +124,8 @@ func Decrypt(src io.Reader, identities ...Identity) (io.Reader, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to read header: %v", err)
 	}
-	if len(hdr.Recipients) > 20 {
-		return nil, errors.New("too many recipients")
+	if len(hdr.Recipients) > 1000 {
+		return nil, errors.New("refusing to decrypt for more than 1000 recipients")
 	}
 
 	var fileKey []byte


ryantm:
CI build failure appears to be unrelated.

```
# filippo.io/age/internal/bech32_test
173internal/bech32/bech32_test.go:46:12: conversion from untyped int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
```

johanfleury:
Why even keep a limit here? There’s not comments in the code and I can’t find a replay of the Twitch stream mentioned in the commit message.

I think, if we keep this limit (either increasing it or keeping it as is), we should document it somehow.

@FiloSottile could you explain the reason behind this check?

ryantm:
The limit does not exist in the specification or the rage implementation.

str4d:
See #139 for continued discussion.

PR #146, opened 2020-10-01
age-keygen: Add option to compute pubkey from privkey

### LINKED ISSUE #122: UX: no way to get public key from private key
there seems to be no way to generate the public key from the private key given the current options available to `age-keygen` and `age`, I was wondering if it was worth having this (if it is possible, i am ignorant in this field) in case you ever lose the "public" part of the key

for example, ssh can do this via 

```plaintext
ssh-keygen -y -f ~/.ssh/id_rsa > ~/.ssh/id_rsa.pub
```

is this useful to add?

dsprenkels:
This PR adds an option to `age-keygen` to compute an Curve25519 public key from the respective private key.

It currently implements this:

```
Usage of ./age-keygen:
  -o FILE
        output to FILE (default stdout)
  -pubkey
        Read the private key from standard input print the corresponding public key.
```

This is different from what I proposed in issue #122, in the way that it *only* reads from standard input. `--pubkey` does not take an (optional) file argument.

The reason is that in order to implement this, we need to overhaul the command line flag parsing, including the usage string. It feels like that is a bit too involved for such a small change.

In the future, we can update `age-keygen` such that it also accepts `-p` (as an alias to `--pubkey`) and that the option can take an optional file argument, from which the private key should be read.

---

Fixes #122

### DIFF
--- a/cmd/age-keygen/keygen.go
+++ b/cmd/age-keygen/keygen.go
@@ -21,6 +21,7 @@ func main() {
 	log.SetFlags(0)
 
 	outFlag := flag.String("o", "", "output to `FILE` (default stdout)")
+	pubkeyFlag := flag.Bool("pubkey", false, "Read the private key file from standard input and print the corresponding public key.")
 	flag.Parse()
 	if len(flag.Args()) != 0 {
 		log.Fatalf("age-keygen takes no arguments")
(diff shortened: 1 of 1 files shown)


FiloSottile:
Thank you for the PR! I went with something a little different, stealing pieces of ssh-keygen and cmd/age, but I liked the idea of using standard input by default!

PR #147, opened 2020-10-05
Return error value when no identity matches a recipient

davidsbond:
Use an error value `ErrNoMatchedIdentities` when `Decrypt` fails due to none of the
provided identities matching the intended recipient. This allows API consumers to specifically
handle the scenario where someone is trying to decrypt something they shouldn't.

This should be a non-breaking API change, as the original error was just an error string.


davidsbond:
Thanks! :)

PR #158, opened 2020-11-26
age: add ParseRecipients

sylr:
Add a ParseRecipients which can read a file containing one public key per line.

It also supports extracting public keys from the output of age-keygen.

Signed-off-by: Sylvain Rabot <sylvain@abstraction.fr>


codesoap:
As I wrote at https://github.com/FiloSottile/age/pull/165#issuecomment-753295900, I think this is not a feature age needs. The `age` tool can be made to use public keys from a file like [testdata/keys2.txt](https://github.com/FiloSottile/age/blob/a495df083beccbea4fa38b0600eb296be3c1335a/testdata/keys2.txt) like this:

```shell
awk 'match($0,/age1.+/) {print "-r", substr($0, RSTART, RLENGTH)}' testdata/keys2.txt | xargs -J % age % -o data.age data
```

I also don't see the need for the new `ParseRecipients()` function, since I feel like this is as too "special"/unintuitive to be part of the official interface. I feel like this should be implemented in a separate library, if needed.

FiloSottile:

PR #167, opened 2020-12-30
armor: add basic benchmarks

codesoap:
I wrote this benchmark, so that I could use `cd armor && go test -bench=Write -cpuprofile out.prof`, because I wanted to find out why encryption with armor is slower than `gpg`. Unfortunately I came to the conclusion, that most time is spend in the base64 package, so I can't do much here and don't need this benchmark anymore.

I create this pull request anyway, in case it is considered useful for potential future investigations.

### DIFF
--- a/armor/armor_bench_test.go
+++ b/armor/armor_bench_test.go
@@ -1,0 +1,58 @@
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+package armor_test
+
+import (
+	"io"
+	"io/ioutil"
+	"math/rand"
+	"strings"
+	"testing"
+
+	"filippo.io/age/armor"
+)
+
+func BenchmarkArmorWrite(b *testing.B) {
+	data := make([]byte, 4096)
+	if _, err := rand.Read(data); err != nil {
+		b.Fatal(err)
+	}
+	armorWriter := armor.NewWriter(ioutil.Discard)
+	defer armorWriter.Close()
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		n, err := armorWriter.Write(data)
+		if err != nil {
+			b.Fatal(err)
+		}
+		b.SetBytes(int64(n))
+	}
+}
+
+func BenchmarkArmorRead(b *testing.B) {
+	fileContents := `-----BEGIN AGE ENCRYPTED FILE-----
+YWdlLWVuY3J5cHRpb24ub3JnL3YxCi0+IFgyNTUxOSB4YWdhZHZ0WG1PZldDT1hD
+K3RPRzFkUlJnWlFBQlUwemtjeXFRMFp6V1VFCnRzZFV3a3Vkd1dSUWw2eEtrRkVv
+SHcvZnp6Q3lqLy9HMkM4ZjUyUGdDZjQKLS0tIDlpVUpuVUQ5YUJyUENFZ0lNSTB2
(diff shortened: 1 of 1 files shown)


FiloSottile:
How much slower than gpg, and how much slower than without armor? I wouldn't be opposed to using an optimised base64 implementation if it's a significant performance boost. The armor is not in the cryptographic hot path so the complexity is relatively cheap.

codesoap:
I tried it with an iso of Xubuntu, that I had at hand:

```console
$ du -h xubuntu-18.04.2-desktop-amd64.iso
1.4G    xubuntu-18.04.2-desktop-amd64.iso
$ time age -r age1df9vthfpykfsd4teaehwau7q8pt0f8d6sad22dgerqamnd43t32qzv54e5 xubuntu-18.04.2-desktop-amd64.iso > /dev/null
age -r age1df9vthfpykfsd4teaehwau7q8pt0f8d6sad22dgerqamnd43t32qzv54e5  >   0.85s user 0.49s system 74% cpu 1.801 total
$ time age -a -r age1df9vthfpykfsd4teaehwau7q8pt0f8d6sad22dgerqamnd43t32qzv54e5 xubuntu-18.04.2-desktop-amd64.iso > /dev/null
age -a -r age1df9vthfpykfsd4teaehwau7q8pt0f8d6sad22dgerqamnd43t32qzv54e5  >   37.87s user 40.49s system 87% cpu 1:30.01 total
$ time gpg2 --compress-algo none -r XXX... -o /dev/null --encrypt xubuntu-18.04.2-desktop-amd64.iso
gpg2 --compress-algo none -r XXX... -o     5.67s user 0.89s system 88% cpu 7.398 total
$ time gpg2 --compress-algo none -r XXX... -o /dev/null --armor --encrypt xubuntu-18.04.2-desktop-amd64.iso
gpg2 --compress-algo none -r XXX... -o      20.55s user 1.16s system 76% cpu 28.465 total
```

codesoap:
I had just measured these times out of curiosity. I don't really see myself encrypting such large files with `-a`, so I would prefer to leave the code as is, instead of introducing an external dependency.

FiloSottile:
Something's off, on my machine (which is arm64 but chacha20poly1305 is pretty good on arm64) `BenchmarkChacha20Poly1305/Seal-8192-8` is `859.92 MB/s` while `BenchmarkEncodeToString-8` is `1377.83 MB/s`. chacha20poly1305 + base64 shouldn't be 50 times slower than chacha20poly1305. I'll look into it.

codesoap:
Thanks for investigating!

I have added `SetBytes()` calls to the benchmarks, which confirms, that the armore Writer and Reader alone are not the problem:

```console
$ go test -bench=.
goos: openbsd
goarch: amd64
pkg: filippo.io/age/armor
BenchmarkArmorWrite-2             152755              8135 ns/op         503.48 MB/s
BenchmarkArmorRead-2              249819              4935 ns/op          73.76 MB/s
PASS
ok      filippo.io/age/armor    3.657s
```

codesoap:
Pefromance looks good now (`age` with armor is now even slightly faster than `gpg` without armor for me):

```console
$ time age -a -r age1df9vthfpykfsd4teaehwau7q8pt0f8d6sad22dgerqamnd43t32qzv54e5 xubuntu-18.04.2-desktop-amd64.iso > /dev/null
age -a -r age1df9vthfpykfsd4teaehwau7q8pt0f8d6sad22dgerqamnd43t32qzv54e5  >   5.46s user 1.74s system 97% cpu 7.389 total
```

Thanks for the fix!

FiloSottile:
Awesome! Thank you for bringing it up ✨

PR #168, opened 2020-12-30
age: add missing -p to usage text

codesoap:
I was initially not sure, whether this flag should only be used for encryption or for both encryption and decryption. This change would have prevented my confusion.


FiloSottile:
Made it a separate variant in #173 because -p and -r are mutually exclusive.

PR #169, opened 2020-12-31
age: create outfile as late as possible

### LINKED ISSUE #159: UX: Empty files created upon passphrase mismatch
## What were you trying to do

Create an encrypted file using a symmetric key. 

## What happened

When a passphrase is entered incorrectly, an empty output file is created.
My expectation was that no file would be created until after the program has performed the encryption operation.

```bash
$ echo "Hello World" > test.txt
$ age -p -o test.txt.age test.txt
Enter passphrase (leave empty to autogenerate a secure one):  Test
Confirm passphrase: test
Error: passphrases didn't match
$ ls -l
-rw-r--r-- 1 user user   12 Nov 29 13:31 test.txt
-rw-r--r-- 1 user user    0 Nov 29 13:32 test.txt.age
```

codesoap:
`age` creates empty output files, when something goes wrong. Examples:
1. With `age -p -o data.age data` when I enter non matching passphrases.
2. With `age -d -o data data.age` when I enter the wrong passphrase.
3. With `age -d -i key -o data data.age` when I provided the wrong secret key.

This pull requests ensures, that the output file is created as late as possible and thus avoids the creation of empty files for the given examples. Unfortunately this leads to some boilerplate, but I feel like it's not too bad.

One drawback is that the user is now informed later about already existing output files. Thus they may have already entered a password, only to be informed that the output file already exists and won't be overwritten. This could be prevented by checking if the file exists in advance. Let me know if you want this implemented.

This pull request should resolve #159 and to some extend #57.

### DIFF
--- a/cmd/age/age.go
+++ b/cmd/age/age.go
@@ -113,7 +113,7 @@ func main() {
 		}
 	}
 
-	var in, out io.ReadWriter = os.Stdin, os.Stdout
+	in := os.Stdin
 	if name := flag.Arg(0); name != "" && name != "-" {
 		f, err := os.Open(name)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


FiloSottile:
Thank you! I took a slightly different implementation strategy, but the examples and discussion was very useful. Let me know if the changes in #173 look good!

codesoap:
I really like the `lazyOpener`. I had briefly thought about something like this, but didn't think it would be implementable with so little code. I'll definitely remember this for future projects.

PR #175, opened 2021-01-03
cmd/age: improve user documentation

codesoap:
Shows where `-R`/`--recipients-file` can be used, adds `-R`/`--recipients-file` to missing-recipient hint, and mention the possibility to use `-` with `-i`.

The `(-r RECIPIENT | -R PATH)...` syntax indicates that either `-r` or `-R` can be used (via `( | )`) and that these arguments may be provided multiple times (via `...`). I believe this is the most common way to represent this in usage texts. I like to take a look at http://docopt.org/ when writing usage texts.


cmd/age/age.go
```diff
@@ -53,9 +53,10 @@ or an SSH public key ("ssh-ed25519 AAAA...", "ssh-rsa AAAA...").
 Recipient files contain one or more recipients, one per line. Empty lines
 and lines starting with "#" are ignored as comments.
```
FiloSottile, line 54:
```suggestion
and lines starting with "#" are ignored as comments. "-" may be used to
read recipients from standard input.
```

cmd/age/age.go
```diff
@@ -32,7 +32,7 @@ func (f *multiFlag) Set(value string) error {
 }
 
 const usage = `Usage:
-    age -r RECIPIENT [-a] [-o OUTPUT] [INPUT]
+    age (-r RECIPIENT | -R PATH)... [-a] [-o OUTPUT] [INPUT]
     age --passphrase [-a] [-o OUTPUT] [INPUT]
     age --decrypt [-i KEY] [-o OUTPUT] [INPUT]
```
FiloSottile, line 37:
```suggestion
    age --decrypt [-i KEY]... [-o OUTPUT] [INPUT]
```

cmd/age/age.go
codesoap: Cool, I didn't even notice that.

PR #176, opened 2021-01-06
cmd/age: Improve decision on when to buffer output

codesoap:
Previously output was buffered when `--armor` was set and the output was a TTY. It was not checked whether the input was a TTY, decryption was not taken into account and binary output was not buffered either. This lead to some problems:

1.: `dd if=/dev/urandom bs=1M count=300 | age -r age1... -a` caused `fatal error: runtime: out of memory` (increase count if it doesn't for you). No one will encrypt this much to stdout, but buffering is unnecessary here.

2.: Output was not buffered for `age -d` (when pasting some armored input into the TTY as input). Admittedly this only happens when copy and pasting a lot of input (ca. 70k bytes for me), but it still seems wrong. Here is how I can make the problem visible:
```
yes 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.' \
| dd bs=1k count=70 \
| age -a -r age1... \
| xclip -i

# Paste with middle mouse button here:
age -d -i mykey
```

3.: Output was not buffered when encrypting and forcing binary output with `age -r age1... -o -`. Of course this is also something most people wouldn't do.

### DIFF
--- a/cmd/age/age.go
+++ b/cmd/age/age.go
@@ -174,13 +174,6 @@ func main() {
 		defer f.Close()
 		out = f
 	} else if terminal.IsTerminal(int(os.Stdout.Fd())) {
-		if armorFlag {
-			// If the output will go to a TTY, and it will be armored, buffer it
-			// up so it doesn't get in the way of typing the input.
-			buf := &bytes.Buffer{}
-			defer func() { io.Copy(os.Stdout, buf) }()
-			out = buf
-		}
 		if name != "-" {
 			if decryptFlag {
 				// TODO: buffer the output and check it's printable.
@@ -190,6 +183,13 @@ func main() {
 				logFatalf("Error: refusing to output binary to the terminal.\n" +
 					`Did you mean to use -a/--armor? Force with "-o -".`)
 			}
+		}
+		if in == os.Stdin && terminal.IsTerminal(int(os.Stdin.Fd())) {
+			// If the input comes from a TTY and output will go to a TTY,
+			// buffer it up so it doesn't get in the way of typing the input.
+			buf := &bytes.Buffer{}
+			defer func() { io.Copy(os.Stdout, buf) }()
+			out = buf
 		}
 	}
 
--- a/cmd/age/age.go
+++ b/cmd/age/age.go
@@ -180,13 +180,16 @@ func main() {
 			buf := &bytes.Buffer{}
 			defer func() { io.Copy(os.Stdout, buf) }()
 			out = buf
-		} else if decryptFlag && name != "-" {
-			// TODO: buffer the output and check it's printable.
-		} else if name != "-" {
-			// If the output wouldn't be armored, refuse to send binary to the
-			// terminal unless explicitly requested with "-o -".
-			logFatalf("Error: refusing to output binary to the terminal.\n" +
-				`Did you mean to use -a/--armor? Force with "-o -".`)
+		}
+		if name != "-" {
+			if decryptFlag {
+				// TODO: buffer the output and check it's printable.
+			} else if !armorFlag  {
+				// If the output wouldn't be armored, refuse to send binary to
+				// the terminal unless explicitly requested with "-o -".
+				logFatalf("Error: refusing to output binary to the terminal.\n" +
+					`Did you mean to use -a/--armor? Force with "-o -".`)
+			}
 		}
 	}
 


str4d:
> Admittedly this only happens when copy and pasting a lot of input (ca. 70k bytes for me), but it still seems wrong.

This is because age chunks plaintext at 64kiB:

- For files smaller than 64kiB, it looks like the output is buffered because it is, by necessity (so that unauthenticated plaintext is never released).
- For large files you'd start to get decrypted output after reading in 64kiB + header size (hence you observing it around 70kiB).

Output is _always_ read in 64kiB chunks, so if there's going to be any buffering at all, it should be a multiple of 64kiB. As for what that should be, it would be useful to know what kinds / sizes of workloads require this buffering; we obviously can't buffer every file in its entirety.

codesoap:
> This is because age chunks plaintext at 64kiB

Sounds plausible to me.

> if there's going to be any buffering at all at decryption time, it should be a multiple of 64kiB

I'm afraid I don't yet understand why it should be a multiple of 64kiB. `bytes.Buffer` simply grows as needed. Why should I limit or complicate this? The buffer is supposed to buffer all output until the full input has been read.

> As for what that should be, it would be useful to know what kinds / sizes of workloads require this buffering; we obviously can't buffer every file in its entirety.

I don't think there are many real use cases and the clipboard probably also has its limit, but using an explicit buffer seems cleaner to me than depending on some default buffer, which probably isn't guaranteed to be stable. Or is this 64kiB buffer a guarantee (if so: could you point me to it's documentation?)?

cmd/age/age.go
```diff
@@ -174,19 +174,23 @@ func main() {
 		defer f.Close()
 		out = f
 	} else if terminal.IsTerminal(int(os.Stdout.Fd())) {
-		if armorFlag {
-			// If the output will go to a TTY, and it will be armored, buffer it
-			// up so it doesn't get in the way of typing the input.
+		if !armorFlag && name != "-" {
+			if decryptFlag {
+				// TODO: buffer the output and check it's printable.
+			} else {
```
FiloSottile, line 180:
```
		if name != "-" {
			if decryptFlag {
				// TODO: buffer the output and check it's printable.
			} else if !armorFlag {
```

`decryptFlag` implies `!armorFlag` so this is the same, but it makes the decision tree clearer.

cmd/age/age.go
```diff
@@ -174,19 +174,23 @@ func main() {
...
+				// If the output wouldn't be armored, refuse to send binary to
+				// the terminal unless explicitly requested with "-o -".
+				logFatalf("Error: refusing to output binary to the terminal.\n" +
+					`Did you mean to use -a/--armor? Force with "-o -".`)
+			}
+		}
+		if terminal.IsTerminal(int(os.Stdin.Fd())) {
+			// If the input comes from a TTY and output will go to a TTY,
+			// and the output will be armored, buffer it up so it doesn't
+			// get in the way of typing the input.
```
FiloSottile, line 190:
```
			// If the input comes from a TTY and output will go to a TTY,
			// buffer it up so it doesn't get in the way of typing the input.
```

Armoring is only half the story, and anyway it's not actually checked in this path, which is fine because who types binary into a TTY.

FiloSottile (commented):
This is better logic, thank you. A couple style comments.

cmd/age/age.go
codesoap: I agree, this looks clearer. I will update the pull request.

cmd/age/age.go
codesoap: Thanks for the hint, I somehow missed this. I'll update the pull request.

cmd/age/age.go
```diff
@@ -174,19 +174,22 @@ func main() {
...
+			if decryptFlag {
+				// TODO: buffer the output and check it's printable.
+			} else if !armorFlag  {
+				// If the output wouldn't be armored, refuse to send binary to
+				// the terminal unless explicitly requested with "-o -".
+				logFatalf("Error: refusing to output binary to the terminal.\n" +
+					`Did you mean to use -a/--armor? Force with "-o -".`)
+			}
+		}
+		if terminal.IsTerminal(int(os.Stdin.Fd())) {
```
FiloSottile, line 187:
```suggestion
		if in == os.Stdin && terminal.IsTerminal(int(os.Stdin.Fd())) {
```

We don't want to buffer output for file inputs, right?

codesoap: Of course! How did I hide so many problems is this little patch 😳

FiloSottile: Hehe, it's messy logic :) Thank you for bringing it up!

PR #177, opened 2021-01-07
cmd/age-keygen: add --output option as an alias for -o

twpayne:
`age-keygen` didn't support the `--output` option it documented -- the option simply wasn't defined. This PR fixes that.

The whitespace change is to make the structure of `cmd/age-keygen.main` more closely match the structure of `cmd/age.main`.


FiloSottile:
Good catch!

PR #183, opened 2021-01-23
Add Fedora installation instructions

eclipseo:
I read an article in [latacora](https://latacora.micro.blog/2019/07/16/the-pgp-problem.html) about GPG shortcomings and it mentioned Age at the end. Since it was not yet packaged or Fedora I decided to add it. It should be available within 10 days after a period in testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,12 @@ On Arch Linux, age is available in the official repositories:
 sudo pacman -Syu age
 ```
 
+On Fedora 33 and later, [`age`](https://src.fedoraproject.org/rpms/age) is available through `dnf`:
+
+```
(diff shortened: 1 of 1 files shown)


alerque:
Great! I suggest letting this PR percolate until there is confirmation that it is available, that way people don't get confused by something that is documented but doesn't work yet.

eclipseo:
@alerque It is now available in Fedora 33 and 34.
(https://bodhi.fedoraproject.org/updates/FEDORA-2021-808c8709d7)

FiloSottile:
Thank you so much! I am very happy to see age packaged.

PR #184, opened 2021-01-29
Bump x/crypto to bring in newer x/sys/unix

qbit:
golang.org/x/sys@v0.0.0-20190412213103-97732733099d is missing support for
OpenBSD/arm64. By bumping x/crypto to the latest version, we can get a x/sys
that works.

Here is the build failure from the OpenBSD build machine:

  http://build-failures.rhaalovely.net/aarch64/2021-01-24/security/age.log


magical:
You can specify indirect dependencies in go.mod, so I think you could just bump the x/sys version, right? Is there any reason to bump x/crypto too?

qbit:
In the past I have encountered upstreams that dislike the idea of adding inderect dependencies. I would prefer to do inderect as it's less invasive.

I will update the pr in a bit if inderect is acceptable.

PR #187, opened 2021-02-13
cmg/age: explicitly set output file in tar example

frankbraun:
On some systems (e.g., OpenBSD) stdout is not the default output for tar. Even on Linux the default if the environment variable TAPE is not set is the compiled-in default (which often is stdout, but doesn't have to be).

This commit changes the example to make it work accross a wider range of systems.


FiloSottile:
I personally use OpenBSD, so I'm sympathetic to its quirks, but people already tend to get confused by `tar` options, and this is already a cluttered example line. We shouldn't make it more complex just to make it work on a niche platform whose users should be well aware of the quirks. After all it's an example, not a cross-platform script.

PR #188, opened 2021-02-16
Add info about Fedora package

mscherer:


mscherer:
Oups, didn't see  #183 was already here.

PR #198, opened 2021-03-14
Add armv7 to age builds

damianb:
armv7 is currently used in the raspberry pi 4 series to an extent; while its hardware supports arm64 (to my knowledge), distribution support is not yet complete.  having armv7 distributions also available will better support low-power hardware like the pi4 better.

as a side-effect, the arm version needs to also be distinguished in the artifact name. resulting artifacts would now be named "armv6" or "armv7" instead of just "arm".

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,6 +30,8 @@ jobs:
             if [ "$GOOS" == "windows" ]; then
               ( cd "$DIR"; zip age.zip -r age )
               mv "$DIR/age.zip" "age-$VERSION-$GOOS-$GOARCH.zip"
+            elif [ "$GOARCH" == "arm" ]; then
+              tar -cvzf "age-$VERSION-$GOOS-${GOARCH}v${GOARM}.tar.gz" -C "$DIR" age
             else
               tar -cvzf "age-$VERSION-$GOOS-$GOARCH.tar.gz" -C "$DIR" age
             fi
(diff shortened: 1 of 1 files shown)


brunobord:
Oh I'd love to see this merged, my raspberry PI is awaiting for me to share secrets with it! ;o)

FiloSottile:
Is there a meaningful performance difference between the armv6 and the armv7 builds?

damianb:
I don't have any particular stats ready offhand. Any suggestions for a computationally stressful operation to use to compare?  I've got a sample size of 4 pi 4Bs in my cluster that I can run against.

FiloSottile:
Encrypting and decrypting a 100MB file with 20 recipients should give a good empiric benchmark. For decryption use the identity for the last recipient, so that they all have to be tried first.

damianb:
Sounds good to me. Let me see what I can throw together. How's a sample size of 100 each sound?

damianb:
Presuming 100 samples, but it's going to work out more like 400; I'm going to run this benchmark against my pi cluster, which is 4 Pi4Bs.

Should work out something like:

--- |pi01|pi02|pi03|pi04
--- | --- | --- | --- | ---
armv6 | 100 | 100 | 100 | 100
armv7 | 100 | 100 | 100 | 100

Variables I really can't account for at this time are going to be:
- RNG exhaustion
- SD card wear
- Manufacturing differences (2 boards were obtained from amazon, two from microcenter and with a delay - probably different batches)
- Thermal throttling

Pretty sure I never dropped haveged on these guys as a direct package, but if it is on there, it's going to be on all of them due to them being collectively managed via ansible.

Probably going to take a bit to pull all the data together, but once I do I'll slap down the analysis and benchmark script together.

age was compiled from commit `56a3f8cce3dc18e60877c6bd04068a585b0922bc` - or the HEAD of this PR's branch, using ubuntu:20.04 running within WSL2.  

Compile command:
```
docker run --rm -v "$PWD":/mnt -w /mnt -e GOOS=linux -e GOARCH=arm -e GOARM=6 golang:1.15 go build -o . ./cmd/...
```

FiloSottile:
Closing for now, but happy to reopen if the benchmarking numbers suggest there is a significant difference.

PR #205, opened 2021-04-05
Use golang.org/x/term instead of deprecated package

codesoap:
According to the [documentation of golang.org/x/crypto/ssh/terminal](https://pkg.go.dev/golang.org/x/crypto/ssh/terminal) this package has been deprecated and golang.org/x/term should be used instead.


FiloSottile:
Thank you!

Issue #196, opened 2021-03-10
Error trying to read both passphrase and input from stdin on Windows


SirRandom:
## Environment

* OS: Windows 10 Pro (x86_64) Build 19041.804
* age version: v1.0.0-rc.1

## What I was trying to do
Not sure if I have the correct cmdline for this but I was trying to password-encrypt some arbitrary words from stdin using
```
age -p -o out
```

## What happened
Seems like age is trying to open `/dev/tty` which does not exist on Windows.

Command line: `age -p -o out`
```
Enter passphrase (leave empty to autogenerate a secure one): Error: could not read passphrase: standard input is not available or not a terminal, and opening /dev/tty failed: open /dev/tty: The system cannot find the path specified.
[ Did age not do what you expected? Could an error be more useful? Tell us: https://filippo.io/age/report ]
```

FiloSottile:
We need to improve the error message, but the issue here is that age is refusing to use stdin for both the passphrase and the input. We could read one line for the passphrase and then consider the rest the input, but it feels like something that's easy to get wrong.

FiloSottile:
Ok, so the logic will have to be a bit more complicated to be safe, since we don't want to read both passphrase and input from a pipe, but doing it from a terminal sounds reasonable, and I think we can make the behavior cross-platform.

PR #258, opened 2021-04-21
if'd out term.IsTerminal/stdinInUse check for Windows

SirRandom:
I added a platform check to the password routine. It's a dirty hack imo but it 'resolves' issue #196 by _wholesale removing_ the error check. On Windows, behavior seems nominal with a password being read from stdin, followed by the input.

This is my first time writing Go and my first pull request. Please let me know if I've done anything horribly wrong. 🙏

### DIFF
--- a/cmd/age/encrypted_keys.go
+++ b/cmd/age/encrypted_keys.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


FiloSottile:
Thank you for the PR! I ended up fixing it a little differently but it was useful to think through a concrete solution.

PR #264, opened 2021-05-03
internal/format: stop NewlineWriter from writing an empty line

### LINKED ISSUE #263: "invalid armor closing line" on inputs with specific lengths
## Environment

* OS: linux
* age version: 1.0.0-rc1

## What were you trying to do

encrypt a thing that's 38 bytes long without having age crash because it emits invalid armored output

## What happened

```
~ % priv=$(age-keygen)
Public key: age1c73ql4n652jz792gwfvyxx53086kyfu95p9t458pqhk8te85yg2qpwcfqg
~ % pub=$(age-keygen -y <(echo $priv))
~ % ct=$(echo "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" | age -e -a -r $pub)                                                          
~ % age -d -i <(echo $priv) <(echo $ct)
Error: invalid armor closing line: 
1 ~ % echo $ct
-----BEGIN AGE ENCRYPTED FILE-----
YWdlLWVuY3J5cHRpb24ub3JnL3YxCi0+IFgyNTUxOSBIZlV5Ly8yL05tYVZ5ZnJG
ZmRhVjlLeVJNK3lhU2oyV1crSmFNeUs2ZkI0CkRtY0RRdW02N0JvZnBXenJOcUZo
MG8yV3hMQVlWR2RXOG5jQTJ6Z2tkOHcKLS0tIGRWdHhJMzZJOXE1OEExbm1wYisr
VGhMaVBhUE1BbUVWSzU0RUgzREFWWW8Kpj8xwx+HyZJOsYUmyDeI328XeMVg2inz
eMFGurvx6Y9ylAQjIV93WwSJqWDVraeIRpzZfXWeF+yPszZvT75GjNuJuZjsoKo=

-----END AGE ENCRYPTED FILE-----
```

codesoap:
I intended to fix #263 with this (and at a first glance it works), but it seems like I'm breaking things, since multiple tests fail now. I couldn't figure out what the problem is. I just wanted to mention my idea for the fix. Feel free to close this PR, if it is silly.

### DIFF
--- a/internal/format/format.go
+++ b/internal/format/format.go
@@ -47,8 +47,7 @@ const BytesPerLine = ColumnsPerLine / 4 * 3
 
 // NewlineWriter returns a Writer that writes to dst, inserting an LF character
 // every ColumnsPerLine bytes. It does not insert a newline neither at the
-// beginning nor at the end of the stream, but it ensures the last line is
-// shorter than ColumnsPerLine, which means it might be empty.
+// beginning nor at the end of the stream.
 func NewlineWriter(dst io.Writer) io.Writer {
 	return &newlineWriter{dst: dst}
 }
@@ -64,6 +63,9 @@ func (w *newlineWriter) Write(p []byte) (int, error) {
 		panic("age: internal error: non-empty newlineWriter.buf")
 	}
 	for len(p) > 0 {
+		if w.written > 0 && w.written%ColumnsPerLine == 0 {
+			w.buf.Write([]byte("\n"))
+		}
 		toWrite := ColumnsPerLine - (w.written % ColumnsPerLine)
 		if toWrite > len(p) {
 			toWrite = len(p)
@@ -71,9 +73,6 @@ func (w *newlineWriter) Write(p []byte) (int, error) {
 		n, _ := w.buf.Write(p[:toWrite])
 		w.written += n
 		p = p[n:]
-		if w.written%ColumnsPerLine == 0 {
-			w.buf.Write([]byte("\n"))
-		}
 	}
 	if _, err := w.buf.WriteTo(w.dst); err != nil {
 		// We always return n = 0 on error because it's hard to work back to the
--- a/internal/format/format.go
+++ b/internal/format/format.go
@@ -165,9 +165,6 @@ func Parse(input io.Reader) (*Header, io.Reader, error) {
 		}
 
 		if bytes.HasPrefix(line, footerPrefix) {
-			if r != nil {
-				return nil, nil, errorf("malformed body line %q: reached footer without previous stanza being closed\nNote: this might be a file encrypted with an old beta version of rage. Use rage to decrypt it.", line)
-			}
 			prefix, args := splitArgs(line)
 			if prefix != string(footerPrefix) || len(args) != 1 {
 				return nil, nil, errorf("malformed closing line: %q", line)
@@ -179,9 +176,6 @@ func Parse(input io.Reader) (*Header, io.Reader, error) {
 			break
 
 		} else if bytes.HasPrefix(line, recipientPrefix) {
-			if r != nil {
-				return nil, nil, errorf("malformed body line %q: new stanza started without previous stanza being closed\nNote: this might be a file encrypted with an old beta version of rage. Use rage to decrypt it.", line)
-			}
 			r = &Stanza{}
 			prefix, args := splitArgs(line)
 			if prefix != string(recipientPrefix) || len(args) < 1 {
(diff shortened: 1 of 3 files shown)


internal/format/format.go
```diff
@@ -47,8 +47,7 @@ const BytesPerLine = ColumnsPerLine / 4 * 3
 
 // NewlineWriter returns a Writer that writes to dst, inserting an LF character
 // every ColumnsPerLine bytes. It does not insert a newline neither at the
-// beginning nor at the end of the stream, but it ensures the last line is
-// shorter than ColumnsPerLine, which means it might be empty.
```
str4d, line 51: This behaviour must not be removed. It was explicitly added (as a breaking change to the format) to ensure that standalone parsing of stanzas will always terminate, which was necessary for the plugin protocol.

codesoap: I'm aware that this is a breaking change, but I was not aware why it was built this way and it seemed unintuitive to me. Could you give me a link to some documentation on the plugin protocol? I can't find anything about plugins in the spec.

str4d: If the last line of a base64 section (like the stanza body) can be a full line, then a streaming parser can't distinguish between the end of a stanza, and an incomplete stanza (due to blocking or errors on IO). This was a problem for the plugin protocol, which is a sequence of stanzas. Requiring the last line to be short removes this ambiguity.

codesoap: Thanks for your answer, but I'm afraid I still don't fully understand the use-case. Why not just wait for the next stanza to begin or end of the header (`--- `)?

str4d: In the header format, you can indeed block header parsing on receipt of the MAC line. In the plugin protocol you can't do that, because you need bidirectional communication, so one direction will pause after a stanza while waiting on the other direction. It either required a non-stanza separator in the plugin protocol (so plugin IPC was no longer just a stream of stanzas), or removal of the ambiguity from parsing; after discussion we opted for the latter.

codesoap: Thanks for the explanation, now I can understand the decision a little better! Is there some place I could read up on this "plugin protocol"? Is it supposed to become part of age/rage/the spec? Is the discussion, that led to your decision publicly visible somewhere?

Although I don't yet know what the plugin protocol is, I assume it is not going to become a fundamental component of age. This assumption reinforces my opinion, that the file format of age should not be complicated in order to cater to some non-primary-use-cases.

FiloSottile:
The plugin protocol will indeed be pretty core to are, because it will be how the CLI and even API is extended without bloating age itself.

However, let's ignore that for a second. "A stanza always ends with a short line" is a simpler rule even in isolation. I actually regret not taking the same approach with the STREAM implementation.

The source of the bug is that I reused the stanza encoding code for the armor. It definitely doesn't make sense for the armor to follow the same rule, but that can be fixed by using PEM rules for it. The armor is not a critical component anyway.

codesoap:
> The plugin protocol will indeed be pretty core

Thanks for the clarification! I'm curiously awaiting it's introduction.

> However, let's ignore that for a second. "A stanza always ends with a short line" is a simpler rule even in isolation.

It is a simple rule indeed and I can see some elegance in it. However, it seems like a unusual rule to me, since I have never seen something like this in a protocol before. An explicit delimiter would seem like a more easily understood choice to me.

I am, for example, used to empty lines being disposable/optional in all plain-text file formats I can think of. I'm aware that the age file format is not to be considered plain text, but I could see myself thinking of the age-file-header as a plain text document, where I may delete a seemingly extraneous empty line in some rare scenario.

Overall, I don't think this is going to cause a lot of problems, but I wanted to share my opinion just to give you my point of view on the decision.

codesoap:
I just had an other idea: Why not skip the line-wrapping in stanzas altogether? This way the first encountered newline character would always signify the end of the stanza. I understand that line-wrapping is a nice-to-have feature when using ascii-armor, but I don't see a good reason for it in the stanzas, because the stanzas are not looked at in the usual use-cases (I think).

internal/format/format.go
FiloSottile: The plugin protocol draft is here: https://hackmd.io/@str4d/age-plugin-spec

I would say that it's going to be the last fundamental component of age, as I expect most other things (YubiKey support, for example) to be built on top of that abstraction.

In general, I think a self-terminating encoding is less complicated per-se, not more. The issue here is that I reused the same code that I used in the internal format for the armor, which honestly might be a mistake.

PR #273, opened 2021-05-31
doc: fix typo

andreaswachowski:


FiloSottile:
Thank you! You can drop the roff and html changes, they are autogenerated files.

andreaswachowski:
Ah, of course, I should have looked closer at the manpage-PR. Done. Thanks for the great tool and hard work on age!

alerque:
> You can drop the roff and html changes, they are autogenerated files.

Why are they tracked in Git if they are autogenerated? Shouldn't they just be part of the build process for where they are needed?

FiloSottile:
Thank you @andreaswachowski! ✨

Issue #128, opened 2020-06-04
opening /dev/tty failed on Powershell and WSL


a-vrma:
## Environment

* OS: Windows 10
* age version: v1.0.0-beta2

## What were you trying to do

Pass a password in without a prompt. Tried on Windows Powershell first, which obviously didn't work because `/dev/tty` doesn't exist. I switched to WSL and it still didn't work, even though `/dev/tty` _does_ work (`echo hi >/dev/tty` works as expected).

## What happened

An error.

```
> echo "pass" | .\age.exe -p -o test.txt.age .\test.txt
Enter passphrase: Error: could not read passphrase: standard input is not available or not a terminal, and opening /dev/tty failed: open /dev/tty: The system cannot find the path specified.
[ Did age not do what you expected? Could an error be more useful? Tell us: https://filippo.io/age/report ]
```

FiloSottile:
The confusing result is because you are running a Windows binary, `age.exe`, from WSL2, which runs in the Windows environment and can't access `/dev/tty` just like when you run it from PowerShell. If you run a Linux binary it runs within the Linux system and works as expected.

```
filippo@dev10:~$ echo test | ./age/age.exe -a -p
Enter passphrase (leave empty to autogenerate a secure one): Error: could not read passphrase: standard input is not a terminal, and opening /dev/tty failed: open /dev/tty: A device which does not exist was specified.
[ Did age not do what you expected? Could an error be more useful? Tell us: https://filippo.io/age/report ]
filippo@dev10:~$ echo test | ./age/age -a -p
Enter passphrase (leave empty to autogenerate a secure one):
Using the autogenerated passphrase "win-steel-arrest-special-property-acid-gym-funny-room-motion".
-----BEGIN AGE ENCRYPTED FILE-----
YWdlLWVuY3J5cHRpb24ub3JnL3YxCi0+IHNjcnlwdCBtZ0hHTmJseWp2UWlMWWt2
R2tPM0JRIDE4CmZ5TEVyL0pZclBIL3RzZkpEQ0VlcHhGOGt5WUZBeE4xWTR5UFNQ
a0xaaWcKLS0tIHZOajVvVk9EeHk0cDZGVGszUk0vNG91UE1vQU45bm9DOG14bFBv
L1pFZkUKvwg8xew1fwyvHsy//XvcetLHoxCYJO+hdkiixs8nHyA3JO7GZQ==
-----END AGE ENCRYPTED FILE-----
```

[...]

PR #274, opened 2021-05-31
cmd/age: use CONIN$ instead of /dev/tty on Windows

codesoap:
I'm intending to solve #128 with this pull request.

The solution for Windows was kindly provided to me by @alexbrainman at https://github.com/golang/go/issues/46164#issuecomment-851329885.

I have to admit, though, that I do not fully understand what is going on, because I am unfamiliar with Windows and it's APIs. As far as I understood using `CONIN$` is also what is [recommended by Microsoft](https://docs.microsoft.com/en-us/windows/console/console-handles):

> The **CreateFile** function enables a process to get a handle to its console's input buffer and active screen buffer, even if `STDIN` and `STDOUT` have been redirected. To open a handle to a console's input buffer, specify the `CONIN$` value in a call to CreateFile. Specify the `CONOUT$` value in a call to **CreateFile** to open a handle to a console's active screen buffer. **CreateFile** enables you to specify the read/write access of the handle that it returns.

### DIFF
--- a/cmd/age/read_password_windows.go
+++ b/cmd/age/read_password_windows.go
@@ -1,0 +1,31 @@
+// Copyright 2021 Google LLC
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+// +build windows
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"golang.org/x/sys/windows"
+	"golang.org/x/term"
+)
+
+func readPassphraseFromTerminal() ([]byte, error) {
+	conin, err := windows.UTF16PtrFromString("CONIN$")
+	if err != nil {
+		return nil, err
+	}
+	tty, err := windows.CreateFile(conin, windows.GENERIC_READ|windows.GENERIC_WRITE, windows.FILE_SHARE_READ, nil, windows.OPEN_EXISTING, 0, 0)
+	if err != nil {
+		return nil, fmt.Errorf("opening CONIN$ failed: %v", err)
+	}
+	defer windows.CloseHandle(tty)
+	defer fmt.Fprintf(os.Stderr, "\n")
+	return term.ReadPassword(int(tty))
+}
--- a/cmd/age/encrypted_keys.go
+++ b/cmd/age/encrypted_keys.go
@@ -46,22 +46,18 @@ func (i *LazyScryptIdentity) Unwrap(stanzas []*age.Stanza) (fileKey []byte, err 
 }
 
 // readPassphrase reads a passphrase from the terminal. If stdin is not
-// connected to a terminal, it tries /dev/tty and fails if that's not available.
-// It does not read from a non-terminal stdin, so it does not check stdinInUse.
+// connected to a terminal, it tries to grab the terminal input and
+// fails if that's not available. It does not read from a non-terminal
+// stdin, so it does not check stdinInUse.
 func readPassphrase() ([]byte, error) {
 	fd := int(os.Stdin.Fd())
 	if !term.IsTerminal(fd) {
-		tty, err := os.Open("/dev/tty")
+		passphrase, err := readPassphraseFromTerminal()
 		if err != nil {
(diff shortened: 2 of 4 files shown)


alexbrainman:
Given that this command is security sensitive, you should have real Windows developer, like @zx2c4, not me review this change.

Alex

FiloSottile:
Awesome, thank you both, this is something that kept biting Windows users! It's kind of unfortunate cmd/age needs to know so much about open terminals. Maybe there's space in x/term for an `OpenTerminal` function.

If @zx2c4 has time to take a look that'd be great, but otherwise I am fairly comfortable with this, the failure mode is not particularly security-sensitive: we don't use this to store a secret, only to read one interactively, so if it doesn't work at most we'll fail to get terminal input. (I guess if permissions are wrong some other user on the system might be able to intercept the input?)

I am getting a decent Windows VM set up so I can test these things properly.

zx2c4:
`CONIN$` is fine and is probably what you want here indeed. What exactly is the security concern here? That it'd read from the wrong console or something like that?

cmd/age/read_password_windows.go
```diff
@@ -0,0 +1,31 @@
...
+	"golang.org/x/sys/windows"
+	"golang.org/x/term"
+)
+
+func readPassphraseFromTerminal() ([]byte, error) {
+	conin, err := windows.UTF16PtrFromString("CONIN$")
+	if err != nil {
+		return nil, err
+	}
+	tty, err := windows.CreateFile(conin, windows.GENERIC_READ|windows.GENERIC_WRITE, windows.FILE_SHARE_READ, nil, windows.OPEN_EXISTING, 0, 0)
```
magical, line 24: I don't think you need x/sys/windows. `os.Open` should work too.

FiloSottile:
If that’s the case we can drop the build tagged files and just check runtime.GOOS to pick the filename.

magical:
Oh my bad, it actually needs to be OpenFile to get r/w permission.

https://play.golang.org/p/FTJwXE0o6AP

Seems to work. I don't know if any of the extra stuff os.OpenFile does
could come back to bite us, but the simplification seems nice.

cmd/age/read_password_windows.go
codesoap: I haven't tried it in `age`, but in another program and it didn't work. See https://github.com/golang/go/issues/46164#issuecomment-851978139 for more info.

magical: @codesoap os.OpenFile works for me (windows 10, go 1.12) 

See https://play.golang.org/p/FTJwXE0o6AP

FiloSottile:
The linked issue mentions Open, which is read only. If you didn't try OpenFile I will this evening.

codesoap:
Sounds good. I haven't tried `OpenFile` yet.

cmd/age/read_password_windows.go
codesoap: That's interesting, thanks. This may even be relevant for https://github.com/golang/go/issues/46164. I'll play around with it a bit as well.

codesoap: I can now confirm that using `os.OpenFile("CONIN$", os.O_RDWR, 0777)` works and `os.OpenFile("CONIN$", os.O_RDONLY, 0777)` does not. Cool find! However, this seems quite odd to me, since `CONIN$` is only supposed to read anyway. Is this to be considered a bug of Windows?!

magical: That's odd, SetConsoleMode says it only needs GENERIC_READ access. It would be nice to know which syscall in term.ReadPassword was throwing the access denied error.

codesoap: If I use the "long way" of reading a password with x/term, we can see, that the error seems to originate from [MakeRaw's implementation on Windows](https://github.com/golang/term/blob/a79de5458b56c188f4fc267a58014ac25fec956a/term_windows.go#L23). So I guess you are correct and `SetConsoleMode` probably is the culprit. Here is a demo: https://play.golang.org/p/qRpuBKj2poq

PR #276, opened 2021-05-31
Add --passphrase-fd option.

sz-coder:
This is my first attempt to get age reading a passphrase from a file descriptor.

See [discussion](https://github.com/FiloSottile/age/discussions/275).

### DIFF
--- a/cmd/age/age.go
+++ b/cmd/age/age.go
@@ -34,7 +34,7 @@ func (f *multiFlag) Set(value string) error {
 
 const usage = `Usage:
     age [--encrypt] (-r RECIPIENT | -R PATH)... [--armor] [-o OUTPUT] [INPUT]
-    age [--encrypt] --passphrase [--armor] [-o OUTPUT] [INPUT]
+    age [--encrypt] --passphrase [--passphrase-fd] [--armor] [-o OUTPUT] [INPUT]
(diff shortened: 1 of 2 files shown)


sz-coder:
Of course there are still some caveats that need to be addressed:

- New line character (0xA) should be trimmed when reading from a file descriptor. (I think?)
- Emit a warning if the passphrase does not fit completely in the buffer (currently 1024 bytes).
- Move logic out of `passphrasePromptForEncryption` function.

FiloSottile:
https://github.com/FiloSottile/age/discussions/275#discussioncomment-867641

clach04:
> ...
> * Compatibility with other operating systems such as windows (I don't know if windows has the concepts of file descriptors).

This would be a non-starter on Windows (but I can see the use for non-Windows).

Cross-referencing related/duplicated discussion/PR topics:

  * https://github.com/FiloSottile/age/discussions/275 "UX: Make age take a passphrase from a non-terminal source"
      * https://github.com/FiloSottile/age/pull/276 "Add --passphrase-fd option."
  * https://github.com/FiloSottile/age/pull/346 "Read passphrase from AGE_PASSPHRASE env var if available"
  * https://github.com/FiloSottile/age/pull/520 "cmd/age: passphrase optionnaly supplied from an environment variable"

PR #316, opened 2021-09-07
fix: release latest age via homebrew

isindir:
Hombrew formula still points to the RC release - fixing.


alerque:
I've opened a PR to put this in Homebrew proper. As soon as any kinks are worked out and it is accepted upstream I suggest just deleting it from this repository. Downstream packaging (at least in such cases where the packaging sources are managed by distro projects anyway) is not really a good fit here at all.

https://github.com/Homebrew/homebrew-core/pull/84805

isindir:
@alerque thanks for the quick response, closing this PR.

PR #328, opened 2021-09-09
Introduce release signing action

technion:
This action allows a fix #326.

This script detects release actions and makes no change to the process on standard push events. You can see it produced a release signed with my key here:

https://github.com/technion/age/releases/download/v2.5/age-v2.5-windows-amd64.zip

You can see here that the signature validates and the code runs on my locked down environment, with no Windows Smartscreen warnings or other difficulties:

image

I've run this script through Shellcheck  - it complains because it doesn't understand the Github variable templating but I made sure no other significant issues were presented.

To operate this action, you will need to setup the following Github secrets for the certificate:

image

I wasn't able to avoid writing the certificate to disk, but I made an effort to delete it as soon as it's used, and it's never accessed in plaintext as the action is designed to require a pass key. It also obtains a timestamp signature so you could consider short term keys and produce long term accepted signatures.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,12 +28,27 @@ jobs:
             curl "https://golang.org/LICENSE?m=text" >> "$DIR/age/LICENSE"
             go build -o "$DIR/age" -ldflags "-X main.Version=$VERSION" ./cmd/...
             if [ "$GOOS" == "windows" ]; then
+              if [ "${{ github.event_name }}" == "release" ]; then
+                echo "Release detected: Signing binaries"
+                echo "${{ secrets.SIGN_CERT }}" >> "$DIR/cert.crt"
+                echo "${{ secrets.SIGN_KEY }}" >> "$DIR/cert.key"
+                /usr/bin/osslsigncode sign -certs "$DIR/cert.crt" -key "$DIR/cert.key" \
+                 -pass "${{ secrets.SIGN_PASS }}" -n "age" -t "http://timestamp.digicert.com" \
+                 -in "$DIR/age/age.exe" -out "$DIR/age/age-signed.exe"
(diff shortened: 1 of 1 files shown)


technion:
It's my understanding the use of an EV certificate is only a requirement for applications that load as drivers. My own certificate was the cheapest one I could find online.

There are unfortunately different grades of "Windows happy", if you look at this workflow:

https://www.theregister.com/2020/06/05/windows_10_microsoft_defender_smartscreen/

Signing it should get it down to one blue popup, and beating that is part an AI signalled on age and and install base.

FiloSottile:
Great, I think I paid and requested a certificate from some cheap reseller. I have approximately no trust in their technical competence, for example because the OrganizationUnit was mandatory and because the minimum key size was 2048 while apparently 3072 is required these days, but maybe they will get me a certificate!

technion:
Many thanks for that!
Unfortunately competence issues aren't limited to cheaper resellers. Consider that all the top end suppliers have statements like this:

>Please use Microsoft Internet Explorer 11 or Mozilla Firefox to collect your digital certificate. Code Signing certificates cannot be generated using Apple Safari, Google Chrome, or Microsoft Edge.

https://sectigo.com/ssl-certificates-tls/code-signing

FiloSottile:
I think it works! Mind trying the binaries in https://github.com/FiloSottile/age/suites/3945491495/artifacts/98935602?

technion:
Looks great to me.
```
> Get-AuthenticodeSignature .\age.exe
    Directory: C:\Users\J\Desktop

SignerCertificate                         Status                                 Path
-----------------                         ------                                 ----
6A3EE16F2306E61831A4D43D488EAECD560C209E  Valid                                  age.exe
```

Runs without warnings.

FiloSottile:
Awesome, merged, thank you!

PR #334, opened 2021-09-25
add mlockall syscall for linux

cnt0:
By default, OS can copy memory pages with passwords, secret keys etc to
the swap file/volume. And swap files are typically not protected, which
means that attacker can easily restore this information.

Mlockall syscall instructs Linux to keep age memory pages in RAM and
never cache them. This eliminates vulnerability described above.

### DIFF
--- a/cmd/age-keygen/keygen.go
+++ b/cmd/age-keygen/keygen.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"filippo.io/age"
+  _ "filippo.io/age/internal/mlockall"
 	"golang.org/x/term"
 )
 
--- a/cmd/age/age.go
+++ b/cmd/age/age.go
@@ -20,6 +20,7 @@ import (
 	"filippo.io/age"
 	"filippo.io/age/agessh"
 	"filippo.io/age/armor"
+  _ "filippo.io/age/internal/mlockall"
 	"golang.org/x/term"
 )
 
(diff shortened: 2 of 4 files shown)


cnt0:
about failed tests. It's required to add `CAP_IPC_LOCK` capability to executables, e.g.
```bash
$ setcap cap_ipc_lock=+ep ./age
```
and I'm not familiar with github pipelines...

alerque:
> It's required to add `CAP_IPC_LOCK` capability to executables

Is that something that needs to be done for regular installations too? If so the setup instructions should also be expanded with this info. Distro packages especially will need to know.

FiloSottile:
Thank you for raising this. Honestly, I am not convinced this is age's responsibility. If your OS or OS configuration leaks arbitrary memory to disk, then all sorts of secrets are at risk: messages, cookies, passwords, file contents... I tend not to agree with the idea that we should guard keys and passwords better than file contents. The latter can be just as sensitive, that's why they are being encrypted after all!

I do like that you CL uses mlockall rather than protect only a few memory regions, but assuming the age output is used _somehow_ (anything but piped to `/dev/null`) then it will be just as much at risk, unless all applications in the chain opt in to the extra protection.

In short, if having plaintext land on disk is not acceptable, you should not have an OS configuration that does that: either disable swap or encrypt it. Requiring any application that could touch sensitive data to mlockall is not going to work. (For example, do cat, tee, and grep call mlockall? If not, wouldn't it be insecure, by the same reasoning, to pipe `age -d` into them?)

PR #341, opened 2021-10-13
Add support for Dependabot

gaby:
* Adds support for keeping dependencies updated with dependabot.


FiloSottile:
Thank you for the PR, but I don't believe in the need to constantly pull up dependency versions.

If there is a security fix, I'll get a notification and assess whether age is impacted. If there is a feature or performance improvement we can pull it in with more purpose. age works well with the current dependencies, and just bringing in changes for the sake of it can only lead to breakage.

Issue #160, opened 2020-12-01
Decryption arguments unexpectedly order dependent


igloo22225:
## Environment

* OS: Linux defiant 5.9.8-arch1-1 #1 SMP PREEMPT Tue, 10 Nov 2020 22:44:11 +0000 x86_64 GNU/Linux
* age version: 6593c56e335f20a76275373a8b1ddf51fb662ab6

## What were you trying to do
Decrypt was a file

## What happened
Decryption arguments unexpectedly order dependent

```
 igloo@defiant  ~/workf  ../git/age/age -d IMG_20201201_205126.jpg.age -i ../git/age/somekey > img.jpg
Error: too many arguments.
age accepts a single optional argument for the input file.
[ Did age not do what you expected? Could an error be more useful? Tell us: https://filippo.io/age/report ]
 ✘ igloo@defiant  ~/workf  ../git/age/age -d -i ../git/age/somekey IMG_20201201_205126.jpg.age > img.jpg 

```

benjojo:
@str4d points out this works in rage

FiloSottile:
We probably don't want to pull in an external flag parsing dependency, but we should detect the issue (by noticing there is a second argument that starts with `-`) and print a suggested fixed command.

PR #345, opened 2021-10-18
Improve cmdline error messages

puenka:
I've added improved error messages and hints for the case `flag.NArg() > 1`.
Assuming that the first entry of `flag.Args` is the file input, I check whether any of the following `os.Args` after the input start with `-`.
In this case, I guess that the user has specified flags after the input (which is not allowed) and therefore suggest to them an alternative command by putting the current file input to the end of the argument chain.
Otherwise - if no flag was found - I guess that the user tried to use multiple file names and print a hint that this is not allowed.

This should solve issue #160.


mikecook:
Rerun build step, it appears Filo fixed the Github secret for osslsigncode certkey pass that failed this build.

mikecook:
rebase on main and it should work now

FiloSottile:
Thank you! I landed a tweaked version of this PR.

PR #353, opened 2021-11-06
refactor: move from io/ioutil to io and os packages

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since age has been upgraded to Go 1.17 in 36b0a4f6117c9ee7210207a69de2eab36b650869, this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

PR #377, opened 2021-12-18
Skip signing if key pass isn't available

mikecook:
Environmental secrets are not shared to workflows run from forks without extra work by Maintainers. 

The existing build workflow uses a secret to sign windows exe's and is breaking for all external PRs. 

Builds are also probably unnecessary for PRs.

See: https://github.com/FiloSottile/age/pull/353#issuecomment-997159334

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,14 +33,16 @@ jobs:
             cp "$RUNNER_TEMP/LICENSE" "$DIR/age"
             go build -o "$DIR/age" -ldflags "-X main.Version=$VERSION" -trimpath ./cmd/...
             if [ "$GOOS" == "windows" ]; then
-              for exe in "$DIR"/age/*.exe; do
-                /usr/bin/osslsigncode sign -t "http://timestamp.comodoca.com" \
-                  -certs .github/workflows/certs/uitacllc.crt \
-                  -key .github/workflows/certs/uitacllc.key \
-                  -pass "${{ secrets.SIGN_PASS }}" \
-                  -n age -in "$exe" -out "$exe.signed"
-                mv "$exe.signed" "$exe"
-              done
+              if [ -n "${{ secrets.SIGN_PASS }}" ]; then
+                for exe in "$DIR"/age/*.exe; do
+                  /usr/bin/osslsigncode sign -t "http://timestamp.comodoca.com" \
+                    -certs .github/workflows/certs/uitacllc.crt \
+                    -key .github/workflows/certs/uitacllc.key \
+                    -pass "${{ secrets.SIGN_PASS }}" \
+                    -n age -in "$exe" -out "$exe.signed"
+                  mv "$exe.signed" "$exe"
+                done
+              fi
               ( cd "$DIR"; zip age.zip -r age )
               mv "$DIR/age.zip" "age-$VERSION-$GOOS-$GOARCH.zip"
             else


.github/workflows/build.yml
```diff
@@ -2,8 +2,6 @@ name: Build and upload binaries
 on:
   release:
     types: [published]
-  push:
```
alerque, line 5: Removing the `push` trigger is more than this commit tries to justify in the commit message & PR details.

alerque (changes requested):
I think build's *should* build for PRs. This is useful for internal development (although whether the developer uses this workflow or not is up to them) to confirm that CI works before a release is tagged. If your builds only run on tagged releases chances are your releases will be broken the first time around when it comes time for them.

Also GitHub has some plumbing to block CI runs on PRs for first time contributors, and for returning contributors having builds in place serves a purpose.

I suggest instead fixing the builds so the signing step is optional and only runs when the source repo is *this* repo and run everything else as normal.

mikecook:
> I think build's should build for PRs. This is useful for internal development (although whether the developer uses this workflow or not is up to them) to confirm that CI works before a release is tagged. If your builds only run on tagged releases chances are your releases will be broken the first time around when it comes time for them.

Reasonable, but it does depend on developer workflow, and that hasn't been established yet.

> Also GitHub has some plumbing to block CI runs on PRs for first time contributors, and for returning contributors having builds in place serves a purpose.

That plumbing is already enabled for this repo, but because environment secrets don't get passed it leads to breakage when secrets are involved.

> I suggest instead fixing the builds so the signing step is optional and only runs when the source repo is this repo and run everything else as normal.

That's my favorite solution, ...and I just figured out how to do that, revising PR

.github/workflows/build.yml
```diff
@@ -9,6 +9,7 @@ permissions:
 jobs:
   build:
     name: Build binaries
+    if: ${{ github.repository == 'FiloSottile/age' }}
```
alerque, line 12: This is closer to what I had in mind, but I would suggest moving this condition farther down so that binaries are always built and only the *signing* step is conditional on the source repo.

alerque (changes requested):

.github/workflows/build.yml
```diff
@@ -60,10 +50,51 @@ jobs:
         with:
           name: age-binaries
           path: age-*
+  sign:
+    name: Sign Windows binaries
+    if: ${{ github.event_name == 'release' }}
```
alerque, line 55: This conditional could still be the "if repository is this canonical one", that way the CI code that does the signing gets a workout and is known to work before a release cycle hits. Also it would make it easier for some people to test intermediary builds.

mikecook: If it's a PR it still will run in canonical but without the secret, so I had to get a little specific with the logic in the newest revision.

.github/workflows/build.yml
```diff
@@ -33,14 +31,6 @@ jobs:
             cp "$RUNNER_TEMP/LICENSE" "$DIR/age"
             go build -o "$DIR/age" -ldflags "-X main.Version=$VERSION" -trimpath ./cmd/...
             if [ "$GOOS" == "windows" ]; then
-              for exe in "$DIR"/age/*.exe; do
```
FiloSottile, line 36: Thank you for working on this!

We could also just add an if around this loop that checks if secrets.SIGN_PASS is empty?

mikecook: I can do that, but it creates an edgecase. If the secret ever gets removed it would silently fail to sign.

mikecook: Which do you prefer?

FiloSottile: That's a valid concern, but I'd rather keep it simple. Failing to sign a binary is not the end of the world. (I would feel differently about the risk of leaking the key, for example.)

alerque (approved):

FiloSottile:
Thank you!

PR #378, opened 2021-12-18
Add Scoop package to README

kotx:
[Scoop](https://scoop.sh/) is a package manager for Windows, similar to Chocolatey. However, Scoop installs as the user instead of as admin, and installs binaries in custom locations such as ~\scoop\apps instead of C:\Program Files etc.


FiloSottile:
Thank you!

PR #381, opened 2021-12-27
Add gomobile support

MarinX:
This PR adds `agemobile` package to support gomobile bind.
Gomobile bind is limited by types and while it allows you to generate identity, you cannot encrypt/decrypt.

With `agemobile` package, we can wrap encrypt/decrypt into supported types (string, bool) so it can be called from Android/iOS.

Attached is an example of a real android device calling the android library(`age.aar`) to generate credentials and encrypt text.
age_mobile_example


FiloSottile:
Hey, this is extremely cool! I think it's a bit out of scope for this repository, also because I don't do gomobile development myself and wouldn't be in a good position to maintain it, but I would be happy to link to it from an upcoming list of age-related projects and integrations if it were a third-party module you host on your GitHub or elsewhere!

paralin:
It's only a couple hundred lines of code, is it worth forking for that really?

FiloSottile:
It would not be a fork, this PR adds a whole separate package which doesn't have to be filippo.io/age/agemobile but can simply be github.com/MarinX/agemobile.

MarinX:
just a fyi
Repo for the mobile library is located here 
[agemobile](https://github.com/MarinX/agemobile)

PR #383, opened 2022-01-04
cmd/age-keygen: errorf calls os.exit

vrnvu:
Shouldn't this function call os.exit as age errorf does?

https://github.com/FiloSottile/age/blob/e4ae4cf884753cbed766b9522902c2a9dd556468/cmd/age/age.go#L409


FiloSottile:
Ah, good catch. It looks like all errorf call sites would thankfully have led to a clean exit or a simple panic if skipped (and not, say, generating an invalid key), so at least it's not a security vulnerability.

FiloSottile:
Introduced in v1.0.0-rc.3 by 0703f86521b047b7a96d168efb72d08e2116b7b6.

PR #385, opened 2022-01-06
age: add --perm flag to set permissions on output file

twpayne:
age currently writes files with permissions 666&^umask, which typically
result in permissions 664 or 644, i.e. world- and group-readable.

If the file should not be world- or group-readable then an extra chmod
command is required, but there is a short window between age writing the
file and updating the permissions.

This commit adds a --perm flag that allows the permissions to be set on
the output file at creation.


FiloSottile:
Thank you, but I don't think this feature belongs in age. One can change the umask to change the permissions of created files, or a different UNIX tool can be used to redirect stdin to a file created with the permissions you like.

twpayne:
Ack, thanks for the fast response :)

PR #386, opened 2022-01-06
age: add --pinentry flag to use pinentry to read passphrase

### LINKED ISSUE #182: Use pinentry if available
rage has `pinentry` support: https://github.com/str4d/rage/pull/64

It would be great if `age` could have it too.

From yubikey-agent it seems that it can be done using: github.com/gopasspw/gopass/pkg/pinentry

https://github.com/FiloSottile/yubikey-agent/blob/4998d8c3d76cbfa97fd1af2953d8eea36ae290e8/main.go

twpayne:
Fixes #182.

This commit uses https://pkg.go.dev/github.com/twpayne/go-pinentry instead of https://pkg.go.dev/github.com/gopasspw/pinentry because the former provides a more complete API, more robust error handling, and correct escape character handling by default.

If this adds too many dependencies, I'd be happy to strip down https://github.com/twpayne/go-pinentry to reduce the dependencies to a minimum.


dominikschulz:
Author of gopasspw/pinentry here. I do support the proposal to replace gopasspw/pinentry with a different implementation. We have introduced our package for an experimental feature that we currently don't put much focus on so our pinentry implementation not our top priority right now (and likely won't be). We kept it around primarily because I knew that yubikey-agent was using it.

If go-pinentry looks sane (IMHO it does) age should probably use that instead. And you should consider adopting it for yubikey-agent, too.

twpayne:
I've updated this to use https://github.com/twpayne/go-pinentry-minimal, which has the same functionality as https://github.com/twpayne/go-pinentry but has no dependencies outside the Go standard library.

FiloSottile:
Thank you @twpayne for the package, and @dominikschulz for chiming in. I've decided we won't merge alternative password input methods in cmd/age, but we'll ship them all in a separate plugin. I'll make sure to support pinentry there and I'll be using the minimal package like in yubikey-agent. See https://github.com/FiloSottile/age/discussions/256#discussioncomment-2638232 for a bit more details.

FiloSottile:
The new batchpass plugin handles non-interactive passphrase encryption. See https://github.com/FiloSottile/age/discussions/256#discussioncomment-15334934 for usage and a warning.

It doesn't include pinentry support, but if you'd like to copy it to make an age-plugin-pinentry, that'd be great!

PR #391, opened 2022-01-20
Host specification in repo instead of Google Docs

paulmillr:
There is zero need to trust Google for hosting. 16 people agreed with me in the corresponding discussion: https://github.com/FiloSottile/age/discussions/228

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1,0 +1,268 @@
+## A simple file encryption tool & format
+
+> Filippo Valsorda (@FiloSottile) — Ben Cartwright-Cox (@Benjojo12)
+
+> Designed at the  [Recurse Center](https://recurse.com/) during NGW 2019
+
+This is a design for a simple file encryption CLI tool, Go library, and format.
+It’s meant to replace the use of gpg for encrypting files, backups, streams, etc.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -0,0 +1,268 @@
+## A simple file encryption tool & format
+
+> Filippo Valsorda (@FiloSottile) — Ben Cartwright-Cox (@Benjojo12)
+
+> Designed at the  [Recurse Center](https://recurse.com/) during NGW 2019
```
waldyrious, line 5: Nit: double space

```suggestion
> Designed at the [Recurse Center](https://recurse.com/) during NGW 2019
```

waldyrious:
Heads up: space characters appear to have been added around links (perhaps as an artifact of the conversion?), which leads to double spaces in various places throughout the doc.

FiloSottile:
Thank you for the PR. I've extracted and polished the most relevant bits of the spec and published it at https://c2sp.org/age, where it will be maintained and versioned separately from the Go implementation.

paulmillr:
Great, finally, thanks

PR #400, opened 2022-03-15
go.mod: Bump dependencies

mrueg:
Since age imports golang.org/x/crypto/ssh, requesting to update deps because of https://github.com/golang/crypto/commit/5770296d904e90f15f38f77dfc2e43fdf5efc083 (not sure if this is exploitable with age, so more applying best practices here)


mrueg:
Thanks for the quick response. I can see your point here. Unfortunately lots of the existing vulnerability scanners are not sophisticated enough yet to detect if it's exploitable or not. 
Thanks also for mentioning `govulncheck` that sounds really interesting!

PR #403, opened 2022-03-24
change Go download link from golang.org to go.dev

mx01xz:
Since Go changed their site from golang.org to go.dev a while ago.


FiloSottile:
Thank you!

PR #408, opened 2022-04-06
Improve Nix install command

MatthewCash:
The current install command for Nix (`nix-env -i age`) is inefficient because it evaluates all of nixpkgs before installing, this is slow and unnecessary.

Changing to `nix-env -f '<nixpkgs>' -iA age` only evaluates the age attribute and is therefore much quicker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ Keep in mind that people might not protect SSH keys long-term, since they are re
     <tr>
         <td>NixOS / Nix</td>
         <td>
-            <code>nix-env -i age</code>
+            <code>nix-env -f '<nixpkgs>' -iA age</code>
         </td>
     </tr>
     <tr>


alerque:
While I don't doubt you are *correct* here, I'm a little bit concerned that the widespread conventional wisdom and hence easily recognizable starter tip. Everywhere on the web I see `nix-env -i $pkgname` and hence when I see it anywhere it is just a tipoff that there is a package and I can "do my usual thing". It seems like the faster alternative is something more advanced Nix-ers would know about anyway and would just be confusing to first-timers. Is that a trade-off worth it for a README on an unrelated project?

FiloSottile:
I agree with Caleb: the age README is not the place to optimize this. If the obvious and simple command is not the best one, that’s a Nix UX issue, not something to fix in every README out there.

MatthewCash:
I understand, and I agree that it is not intuitive to have two ways to install packages, especially since it is not immediately clear the difference between the two methods when looking at the commands.
The real concern I had is that the command will likely continue to get slower as nixpkgs gets bigger, but you are right that this is fundamentally a Nix issue and if they can't even be consistent in their own manual, then it is probably too much to ask everyone else to fix it individually.

PR #409, opened 2022-04-20
Fix CVE-2022-27191

deryoman:
Update x/crypto to latest version to fix [CVE-2022-27191](https://nvd.nist.gov/vuln/detail/CVE-2022-27191) in age and age-keygen

- https://access.redhat.com/security/cve/CVE-2022-27191
- https://github.com/advisories/GHSA-8c26-wmh5-6g9v
- https://groups.google.com/g/golang-announce
- https://groups.google.com/g/golang-announce/c/-cp44ypCT5s


deryoman:
Hey there,
unfortunately I do not have a FreeBSD box at hand to test. The failure seems to be infrastructure related, the FreeBSD runner is apparently missing the git executable:
```
go: missing Git command. See https://golang.org/s/gogetcmd
```
Could anybody jump in here? :)

FiloSottile:
Hi, CVE-2022-27191 does not affect age at all. It's a crash in the SSH server when using custom Signers. We only use key serialization and deserialization from golang.org/x/crypto/ssh. Vulnerability scanners that indiscriminately flag any vulnerability in a dependency are a problem, because they reduce the signal to noise ratio, and take attention away from actual vulnerabilities, training projects to just blindly merge fixes without doing the research of "how does this impact my users and should we tell them".

PR #412, opened 2022-04-28
docs: added alpine package to README

tionis:
Alpine Linux also has age packaged, as can be seen here:
https://pkgs.alpinelinux.org/packages?name=age&branch=edge
I just added the install command to the README.


FiloSottile:
Nice, thanks!

Issue #279, opened 2021-06-02
Add scripting examples


FiloSottile:
Including how to use an identity stored somewhere like `pass`, and generating a new key pair and doing something with the recipient line. We already have an example for how to script sending to a GitHub user.

This could also encourage `-e` usage for explicit scripts, and maybe show how to use `-e -i`.

FiloSottile:
This can also mention that we don't support scripting passphrases and offer alternatives, like a passphrase-encrypted identity file.

dani-CO-CN:
I needed a way to decrypt an age-encrypted identity file. I had quite a hard time doing that in python without user input, i ended up using `expect` or the python package `pexpect`.

```python
import pexpect
pexpect.run('/path/to/age -d /path/to/age.key', events={'Enter passphrase:': 'SomePassword\n'})
```
Otherwise fiddling with `/dev/tty` and file descriptors in python sub-processes seemed even more involved.

brandsimon:
I want to use age to decrypt a file in an initramfs, where /dev/tty is not available.
```
could not read passphrase: open /dev/tty: no such file or address
```
I also thought about using expect, but I dont want to pack it in the initramfs. Using a pipe or anything does not work.
Are there other ways than to use expect?

brandsimon:
I use it in a initramfs with dracat and mkinitcpio, it works just fine: https://gitlab.com/cryptographic_id/cryptographic-id-rs/-/blob/main/usr/lib/cryptographic_id/show_identities?ref_type=heads#L27
My problem got fixed here: https://github.com/FiloSottile/age/commit/ac31f5c9356f42c3ec76440bff74ae4bac9de794

brandsimon:
@dkwo No big ones, but I have no experience with yubikey in initramfs. You need the device drivers in the initramfs, probably some configuration files and maybe some udev rules.

PR #414, opened 2022-05-03
cmd/age: also fallback to stdin and stdout, if /dev/tty exists, but cannot be opened

brandsimon:
As mention here [0], I had a problem when /dev/tty exists, but age failed to open it. Now age falls back to /dev/tty, which is the same behavior as if /dev/tty does not exist.
If there are additional changes needed, I am happy to implement them.

PS: Thank you very much for this great tool!

[0] https://github.com/FiloSottile/age/issues/279#issuecomment-1115499664

### DIFF
--- a/cmd/age/encrypted_keys.go
+++ b/cmd/age/encrypted_keys.go
@@ -119,11 +119,7 @@ func readPassphrase(prompt string) ([]byte, error) {
 			return nil, err
 		}
 		defer out.Close()
-	} else if _, err := os.Stat("/dev/tty"); err == nil {
-		tty, err := os.OpenFile("/dev/tty", os.O_RDWR, 0)
-		if err != nil {
-			return nil, err
-		}
+	} else if tty, err := os.OpenFile("/dev/tty", os.O_RDWR, 0); err == nil {
 		defer tty.Close()
 		in, out = tty, tty
 	} else {


cmd/age/encrypted_keys.go
```diff
@@ -119,8 +127,7 @@ func readPassphrase(prompt string) ([]byte, error) {
 			return nil, err
 		}
 		defer out.Close()
-	} else if _, err := os.Stat("/dev/tty"); err == nil {
-		tty, err := os.OpenFile("/dev/tty", os.O_RDWR, 0)
+	} else if tty, err := openTTY(); err == nil {
```
FiloSottile, line 130:
```suggestion
	} else if tty, err := os.OpenFile("/dev/tty", os.O_RDWR, 0); err == nil {
```

No need for the Stat at all, right?

brandsimon: Totally agree, good catch, thank you!

cmd/age/encrypted_keys.go
```diff
@@ -119,8 +119,7 @@ func readPassphrase(prompt string) ([]byte, error) {
 			return nil, err
 		}
 		defer out.Close()
-	} else if _, err := os.Stat("/dev/tty"); err == nil {
-		tty, err := os.OpenFile("/dev/tty", os.O_RDWR, 0)
+	} else if tty, err := os.OpenFile("/dev/tty", os.O_RDWR, 0); err == nil {
 		if err != nil {
```
FiloSottile, line 123: No need for this if anymore.

brandsimon: Good catch! Thank you

FiloSottile:
Thank you!

PR #422, opened 2022-05-21
feat: move internal to plugin

tomberek:
This is not pretty, but was a very rough attempt to see if it would work to expose the [plugin system to sops](https://github.com/tomberek/sops/commit/a4849fcd145656c228860b0939e5b228811789c9)

It results in a sops binary capable of age1yubikey/AGE-PLUGIN-YUBIKEY plugin actions. So that is promising. Is there continued work to do something along these lines?


FiloSottile:
Hey! Yeah, the plan is to expose this API, indeed. Seeing it used is extremely helpful, thank you!

Looks like besides the current internal plugin package, you'd want a function to parse both native and plugin recipients and identities, and a TUI implementation of the callbacks, is that right? Does the age TUI work as-is for SOPS?

tomberek:
Yes! Surprisingly these two sets of changes allowed me to use sops + age + age-plugin-yubikey. (warning: I did extremely minimal testing.)

FiloSottile:
Thank you for experimenting with this! I plan to expose the plugin API after some cleanup. Closing this because the filippo/plugins branch is being deleted.

tomberek:
Is it happening? https://github.com/FiloSottile/age/commit/6976c5fca5d828b3e74e3a9721583af5478e8244

FiloSottile:
Hi! Yeah you should find everything but the TUI in https://pkg.go.dev/filippo.io/age/plugin@v1.1.2-0.20230805193447-c89f0b932ecd. Let me know if it looks good, ideally before the next release :)

Kranzes:
Awesome! So here's a question, when are we getting a release with this? :wink:

PR #427, opened 2022-06-20
Update minimum Go version

mohammed90:
The commit 4f7bb44 uses the fuzz testing API of stdlib, which was introduced in go1.18. The declared minimum Go version is 1.17, which precludes the fuzzing API. Update the minimum Go version to 1.18 to accommodate the needs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: [1.16.x, 1.17.x]
+        go: [1.17.x, 1.18.x]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: [1.16.x, 1.17.x]
+        go: [1.17.x, 1.18.x]
```
mohammed90, line 11: Oh, I guess this means 1.17 should be dropped entirely. @FiloSottile, are you ok with this?

FiloSottile:
Hi! Thank you for the PR. age works well down to Go 1.16 (which is the `go mod tidy` compatibility target when the `go.mod` version is Go 1.17). There are a few new tests that require Go 1.18, but I just build tagged those, so they won't break the other tests.

PR #455, opened 2022-10-09
.github/workflows: use gh to upload release artifacts

0x2b3bfa0:
Uses [`gh release upload`](https://cli.github.com/manual/gh_release_upload) for the sake of pipeline code golfing, if that's a thing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,26 +75,6 @@ jobs:
         with:
           name: age-binaries
       - name: Upload release artifacts
-        uses: actions/github-script@v3
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          script: |
-            const fs = require("fs").promises;
-            const { repo: { owner, repo }, sha } = context;
-
-            const release = await github.repos.getReleaseByTag({
-              owner, repo,
-              tag: process.env.GITHUB_REF.replace("refs/tags/", ""),
-            });
-            console.log("Release:", { release });
-
-            for (let file of await fs.readdir(".")) {
-              if (!file.startsWith("age-")) continue;
-              console.log("Uploading", file);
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           path: age-*
   upload:
     name: Upload release binaries
-    if: ${{ github.event_name == 'release' }}
+    if: github.event_name == 'release'
```
0x2b3bfa0, line 67: This suggestion deviates from the pull request topic

FiloSottile:
Oh, thanks, this is definitely better. Why `GH_REPO` and why `GH_TOKEN` instead of `GITHUB_TOKEN` as documented in https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows.

0x2b3bfa0:
> _Why `GH_REPO`_

Because the `upload` job does not use the [`checkout`](https://github.com/actions/checkout) action. Thus, when `gh` is run, it won't find a Git repository nor know where to upload the artifacts.

```console
$ git release upload ...
fatal: not a git repository (or any of the parent directories): .git
/usr/local/bin/git: exit status 128
```

The easiest ways of specifying the target repository are either specifying the `--repo` option or setting the [`GH_REPO`](https://cli.github.com/manual/gh_help_environment) environment variable.

> _Why [`GH_TOKEN`](https://cli.github.com/manual/gh_help_environment) instead of `GITHUB_TOKEN`_

Just for foolish consistency with the `GH_REPO` environment variable; feel free to use `GITHUB_TOKEN` if you deem it more appropriate. The same applies for `github.token` versus `secrets.GITHUB_TOKEN` to retrieve it.

FiloSottile:
That all makes sense, thank you!

PR #456, opened 2022-10-09
.github/workflows: parallelize build process

0x2b3bfa0:
Speeds up the build process and simplifies the build script a little bit.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,16 @@ jobs:
     name: Build binaries
     runs-on: ubuntu-latest
     environment: "Build, sign, release binaries"
+    strategy:
+      matrix:
+        include:
+          - {os: linux, arch: amd64}
+          - {os: linux, arch: arm, arm: 6}
+          - {os: linux, arch: arm64}
+          - {os: darwin, arch: amd64}
+          - {os: darwin, arch: arm64}
+          - {os: windows, arch: amd64}
+          - {os: freebsd, arch: amd64}
     steps:
       - name: Install Go
         uses: actions/setup-go@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -22,41 +32,36 @@ jobs:
           fetch-depth: 0
       - name: Build binaries
         run: |
-          sudo apt-get update && sudo apt-get install -y osslsigncode
```
0x2b3bfa0, line 25: Moved under the conditional so it's installed only when `matrix.os` is `windows`

.github/workflows/build.yml
```diff
@@ -20,43 +30,38 @@ jobs:
...
+          else
+            tar -cvzf "age-$VERSION-$GOOS-$GOARCH.tar.gz" -C "$DIR" age
+          fi
+        env:
+          CGO_ENABLED: 0
+          GOARCH: ${{ matrix.arch }}
+          GOARM: ${{ matrix.arm }}
+          GOOS: ${{ matrix.os }}
       - name: Upload workflow artifacts
         uses: actions/upload-artifact@v2
```
FiloSottile, line 66: So does `actions/upload-artifact@v2` combine multiple uploads?

.github/workflows/build.yml
```diff
@@ -11,6 +11,16 @@ jobs:
...
+    strategy:
+      matrix:
+        include:
+          - {os: linux, arch: amd64}
+          - {os: linux, arch: arm, arm: 6}
+          - {os: linux, arch: arm64}
+          - {os: darwin, arch: amd64}
+          - {os: darwin, arch: arm64}
+          - {os: windows, arch: amd64}
+          - {os: freebsd, arch: amd64}
```
FiloSottile, line 23: As a matter of style, let's keep calling these GOOS/GOARCH/GOARM.

.github/workflows/build.yml
0x2b3bfa0: Yes, all the artifacts will be appanded under the same `name`

.github/workflows/build.yml
0x2b3bfa0: Resolved with https://github.com/FiloSottile/age/pull/456/commits/d9f5a6c7d08cb48a5ebf53da3c11b44ea08085df 🙃

.github/workflows/build.yml
0x2b3bfa0: #### Minimal reproducible example

```yaml
on: workflow_dispatch
jobs:
  create:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        file: [1, 2]
    steps:
    - run: date > file-${{ matrix.file }}
    - uses: actions/upload-artifact@v2
      with:
        name: files
        path: file-*
  read:
    needs: create
    runs-on: ubuntu-latest
    steps:
    - uses: actions/download-artifact@v2
      with:
        name: files
    - run: ls

```

0x2b3bfa0: (Side note: another nice preoperty of artifacts is that you can also overwrite them)

FiloSottile:
Thank you!

PR #458, opened 2022-10-26
Bump golang.org/x/sys to newer version

joaopapereira:
During a trivy scan, a vulnerability was found https://avd.aquasec.com/nvd/2022/cve-2022-29526/ in `age` library. 
This doesn't look like it impacts the library itself but we are bumping this dependency to ensure we do not have false negatives.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.17
 require (
 	filippo.io/edwards25519 v1.0.0-rc.1
 	golang.org/x/crypto v0.0.0-20210817164053-32db794688a5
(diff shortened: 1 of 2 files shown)


alerque:
What is a "false negative" and how does bumping a dependency that you already determined does *not* effect this library make it more secure?

FiloSottile:
age is not affected by CVE-2022-29526.

My firm position is that upgrading everything that depends on a widely used module every time it publishes a vulnerability affecting a small subset of users is counter-productive, as it discourages publishing vulnerabilities and generates busywork.

PR #465, opened 2022-12-13
Better passphrase generation

luizsdf:
Generate 12 words instead of 10 for 128-bit entropy (`log(2048, 2) * 12`) and replace the separator `'-'` with `' '` for simplicity.


FiloSottile:
Hello! Thank you but these were deliberate design choices. 110 bits of entropy are plenty, especially when stretched by scrypt. Dashes are used to make the password easily copy-pastable and to make it less likely users will mistakenly leave out a word.

PR #475, opened 2022-12-26
cmd/age: fix terminal escape sequences on Windows

### LINKED ISSUE #474: Weird characters in place of terminal escape sequences on Windows Console
## Environment

* OS: Windows 10 Pro
* age version: v1.1.1

## What were you trying to do

I was trying to encrypt a file with a passphrase.

## What happened

On both Windows PowerShell and cmd.exe, terminal escape sequences do not work by default. As a result, I obtained the following output, containing weird characters on screen.

age-escape-sequences-issue-powershell
age-escape-sequences-issue-cmd

As a side note, I actually discovered this issue while trying to use the same functions as yours to ask for a passphrase. I strongly believe that this behavior is present since the introduction of escape sequences in 349ed5ed3fcee75238ea83c59c38efedc3bfccae. This commit post-dates age version v1.0.0, so it may be the case that not many people have encountered this issue, but I think that it would negatively impact the UX for Windows users from now on (indeed, v1.1.0 and v1.1.1 just came out).

I'm going to make a PR to address this issue as a follow-up.

P.S.: sorry for not disclosing this issue before the v1.1.0 release, I didn't see it coming!

BuriedInTheGround:
With this PR, if possible, we enable virtual terminal processing, which is necessary for using terminal escape sequences on instances of the Windows Console. When enabling virtual terminal processing fails, we completely avoid using escape sequences to prevent weird characters to be printed to the console.

Please see comments on code for more info.

Fixes #474

P.S.: I'm happy to explain eventually obscure choices I made.


cmd/age/tui_windows.go
```diff
@@ -0,0 +1,45 @@
...
+	err := syscall.GetConsoleMode(syscall.Stdout, &mode)
+	if err != nil {
+		// Terminal escape sequences may work at this point, but we can't know.
+		avoidTerminalEscapeSequences = true
+		return
+	}
+
+	mode |= ENABLE_PROCESSED_OUTPUT
+	mode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING
+	ret, _, _ := setConsoleMode.Call(uintptr(syscall.Stdout), uintptr(mode))
```
FiloSottile, line 39: Thank you for this and sorry about the review lag!

I made a couple small style changes, but then realized this shouldn't be unconditionally be stdout, as we might have opened `CONOUT$` as a terminal instead.

BuriedInTheGround: Thanks for taking the time to review!

Shouldn't this work anyway by using `syscall.Stdout`? Go initializes `syscall.Stdout` with `GetStdHandle(STD_OUTPUT_HANDLE)` (see [here](https://pkg.go.dev/syscall?GOOS=windows#Stdout)), and `STD_OUTPUT_HANDLE` defaults to `CONOUT$`, as described in the [Windows Console docs](https://learn.microsoft.com/en-us/windows/console/getstdhandle).

Maybe I'm confusing or missing something. If so, could you please elaborate on the situation?

FiloSottile: That won't work if stdout is being redirected.

> The standard handles of a process may be redirected by a call to [SetStdHandle](https://learn.microsoft.com/en-us/windows/console/setstdhandle), in which case GetStdHandle returns the redirected handle. If the standard handles have been redirected, you can specify the CONIN$ value in a call to the [CreateFile](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea) function to get a handle to a console's input buffer. Similarly, you can specify the CONOUT$ value to get a handle to a console's active screen buffer.

I have amended the commit to use withTerminal's out file handle.

PR #476, opened 2022-12-29
age: depend on c2sp.org/CCTV/age

bdd:
Makes it easier for package builders with no networking (e.g Nixpkgs) to vendor modules and run tests without further need of network access.

Will update the pseudo versions in this PR once https://github.com/C2SP/CCTV/pull/2 merges.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 
 // Test dependencies.
 require (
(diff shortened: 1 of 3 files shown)


FiloSottile:
Thank you! ✨

PR #481, opened 2023-01-06
.github/workflows: trigger interoperability tests using gh

0x2b3bfa0:
After cli/cli#6614, `gh api` can be used with nested object fields. 💅🏼 [^1]

> **Note** includes 79cc0644c2eacabb212740ec3ef11290220b4c01 with a pedantic change that doesn’t provide any security benefit on a public repository; can be dropped upon request.

[^1]: 🟩

### DIFF
--- a/.github/workflows/interop.yml
+++ b/.github/workflows/interop.yml
@@ -8,8 +8,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Trigger interoperability tests in str4d/rage
-        run: |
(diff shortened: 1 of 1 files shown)


FiloSottile:
Thank you! Much better :)

I dropped 79cc0644c2eacabb212740ec3ef11290220b4c01 because I find it a little less explicit. ("Does `{}` mean no permissions or is it ignored? It should be the former, but who knows with the GitHub permission configs? Let's check the docs...")

PR #482, opened 2023-01-11
Fix apostrophes in age(1)

zhsj:
Apostrophes should be entered as ' ; using \' produces an acute accent.

Co-authored-by: Shengjing Zhu <zhsj@debian.org>

This reworks [patch](https://salsa.debian.org/go-team/packages/age/-/blob/debian/1.0.0-2/debian/patches/0002-Fix-apostrophes-in-age-1.patch) from @ryanakca on latest version.


FiloSottile:
Groff files are generated by the relative Ronn files. Please update those instead. (CI will regenerate the Groff and HTML.)

zhsj:
Oops, I didn't notice it's generated. So it problem from ronn, which generates `\'` for `'`.

PR #483, opened 2023-01-11
.github/workflows: update ronn to 0.9

zhsj:


FiloSottile:
Thank you! Less complexity this way.

PR #511, opened 2023-06-28
plugin: return error var

quite:
For convenient going through multiple identities and finding which ones we can handle.


quite:
hey @FiloSottile i found that this would be convenient when i'm implementing the convert (-y) function for age-plugin-tkey

quite:
Replaced by https://github.com/FiloSottile/age/pull/518 which targets main.

PR #513, opened 2023-06-30
Fix misplaced ) in README.md

dilbernd:


FiloSottile:
No, the current placement is correct. `-` is an argument to `-R` or `-i`.

dilbernd:
Sorry

Issue #142, opened 2020-09-07
Support for ecdsa-sha2-nistp256 SSH public key algorithm


fzakaria:
## Environment

* OS: Linux

## What were you trying to do

> Encrypt a file with my public key using **ecdsa-sha2-nistp256**

I was following a workflow where I wanted to encrypt a secret with all public keys listed on my GitHub account -- https://github.com/fzakaria.keys
One of the keys is _ecdsa-sha2-nistp256_ which age does not seem to support.

## What happened

```
echo "hello world" | age -r "ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBMSpLrw3iBKaZTxtOY6E8xzekgmhqjnCKl+biR7/AxIqkB2aqK+rO3F3p2oD3SQvJJuw/ifmryq8FOjlOH1JQ6M=" -a
Error: unknown recipient type: "ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBMSpLrw3iBKaZTxtOY6E8xzekgmhqjnCKl+biR7/AxIqkB2aqK+rO3F3p2oD3SQvJJuw/ifmryq8FOjlOH1JQ6M="
[ Did age not do what you expected? Could an error be more useful? Tell us: https://filippo.io/age/report ]
```

joonas-fi:
One pro about supporting P256 is that there are HSMs like the YubiKey which allow you to do ECDH inside a trusted environment, so one could do ephemeral-static ECDH for P256 (like the existing X25519 approach is using). https://github.com/go-piv/piv-go/ supports this (thanks to @tv42)

Currently there's no ubiquitous HSM (like the YubiKey) with support to Curve25519, so P256/P384 is the best we could have with hardware protection support. P256 seems to be about an order of magnitude faster, though.

andir:
From my limited understanding of the matter you could also use the "Secure Enclave" on Mac Books to store the key material and perform operations just like https://github.com/sekey/sekey or https://github.com/maxgoedjen/secretive (and potentially many more).

FiloSottile:
We'll offer YubiKey (PIV) support as a plugin soon, but it will not be through SSH keys, it will be a dedicated format.

[...]

PR #522, opened 2023-09-01
Adding support for ECDSA NIST curve SSH keys

fakegermano:
As they are becoming more common in the field, This PR adds support for the use of that type of key for encryption/decryption.

This PR also makes use of the new go 1.21 `ecdh` package that natively implements the key exchange for elliptic curves (supports all `X25519` and` P256`, `P384` and `P521`). 

I can also add similar PRs for the `rust` and `typescript` implementations when I have time, but let me know if I have missed something or messed something up, or If I need to add more tests :)


fakegermano:
as per this: https://github.com/FiloSottile/age/issues/142 (which I discovered very late I know) I will close this MR, as it seems we do not intend to add the feature for security reasons.

PR #564, opened 2024-04-20
Fix scoop command in README.md

rasa:


gaby:
@rasa This is not fixing anything, the command was correct already. With the & it will run the 2nd command regardless of exit status of the first one.

rasa:
> @rasa This is not fixing anything, the command was correct already. With the & it will run the 2nd command regardless of exit status of the first one.

@gaby Then use `&&`. The `;` char does not separate characters in a CMD prompt window.

PR #579, opened 2024-06-16
Bump various dependencies

FiloSottile:


.github/workflows/test.yml
```diff
@@ -30,6 +30,10 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Install bootstrap Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.22
```
mohammed90, line 36: You can use `latest` if you'd like a default usable value

Issue #526, opened 2023-10-27
Plugin receives only one identity or recipient when there are multiple


olastor:
## Environment

* OS: Linux (Fedora 37)
* age version: 1.1.1

## What were you trying to do

Creating a plugin according to the spec.

## What happened

When providing multiple recipients, e.g.,

```
cat secret.txt | age -e -r age1my-plugin<recipient 1 data> -r age1my-plugin<recipient 2 data> -o secret.txt.enc
```

or doing the same via an identity file with two identities in it (one per line): 

```
cat secret.txt | age -e -i ids.txt -o secret.txt.enc
```

the controller (age) only sends one recipient (or identity), e.g.,

```
-> add-identity AGE-PLUGIN-MY-PLUGIN<identity 1 data>

-> wrap-file-key
[...]
-> done
```

Checking the processes during encryption, it seems that **age creates a new process for the second identity after it wrapped to the first identity**, so each instance of the plugin seems to only always gets to see one recipient/identity. That makes it hard to implement the [spec](https://github.com/C2SP/C2SP/blob/main/age-plugin.md#phase-1-client-uni-directional) correctly:

> Having assembled the full set of recipients and identities that the client wishes to wrap to, the plugin determines whether it can successfully wrap to all recipients and identities. The plugin MUST generate an error if one or more recipients or identities cannot be wrapped to.

It's also worth noting that for decryption also only one identity is sent to the plugin. In the case of encryption, it's not so bad because the key needs to be wrapped to all recipients/identities anyways (it has some other side-effects in my case though). But for decryption this limits the plugin because it is not able to choose which among all recipients/identities to use for unwrapping (only needs to happen for one of them).

Using `rage`, this seems to work differently, i.e. there's one plugin process that can index all identities/recipients:

```
-> add-identity AGE-PLUGIN-MY-PLUGIN<identity 1 data>

-> add-identity AGE-PLUGIN-MY-PLUGIN<identity 2 data>

-> wrap-file-key
[...]
-> done
```

Although for `rage` I also opened a similar issue for the decryption phase: https://github.com/str4d/rage/issues/414

olastor:
It seems this is intentional: https://github.com/str4d/rage/issues/414#issuecomment-1925463164

PR #580, opened 2024-06-17
plugin: new framework to implement plugins

FiloSottile:


AnomalRoil (commented):
I'm not entirely done with my tests but I can confirm both the standard and the interactive mode worked for me.

AnomalRoil:
Okay, a first set of issues when trying to use `plugin` when using `age` as a library willing to support arbitrary age plugin recipients:
- the currently exposed `age.ParseRecipient` and `age.ParseIdentity` are both only returning ones of type `*X25519Identity` only.
- there exist a `parseIdentities` function supporting plugin ones, but it seems the only way to use it is to rely on `EncryptedIdentities` and there's no such things for recipients.

The rest of the library code is fairly agnostic of whether this is a plugin recipient or identity, so having an easy way of parsing plugin identities would be nice. 

Currently #518 is providing a way to detect "regular" identities from plugin identities through error handling, which could work as well but isn't as convenient I guess, although it's more in line with the rest of the currently exposed parsing functions (one is supposed to parse SSH vs age recipients and identities using the specific exported functions for each currently, meaning everyone ends up re-implementing [this](https://github.com/FiloSottile/age/blob/faefdc3c81efa89fd41998b77c62234ca56be10b/cmd/age/parse.go#L31-L45)).

In other notes:
- the `plugin.NewIdentity` and `plugin.NewRecipient` are both expecting a `plugin.ClientUI` but there isn't any exported "default" ClientUI. This might be a decision to force people to implement their own?
- The existing `plugin.ClientUI` interface does not expect any context to be passed to any of the prompts, whereas a library would be typically carrying these around, especially for such prompts that we might want to timeout on. It might be good to refactor the `ClientUI` interface to expect contexts in each of the function calls?

olastor:
It works great in my tests so far :tada: 

One thing I noticed though is that the age client that communicates with the plugins currently still seems to spawn separate processes. For example, `age -d -i identities.txt -o test somefile.enc` (two identities in `identities.txt`) will do something like this:

```
-> add-identity AGE-PLUGIN-SSS-1R79SSQQQQQQQQQ8LQNQVZZKZXQX
-> grease-1331f5cec305b37b

-> recipient-stanza 0 X25519 DsW/n
xxxx
-> done

-> msg
SGFuZGxlZCBpZGVudGl0eSBpbiAxNzEzNQ
age: sss plugin: Handled identity in 17135
-> ok

-> grease-7cada51a126db013 78
gBr1Q7wtAl4/a8vPQe/OLdZyc0z5hVVuf06OuIxigh7vNBlTn07r8PTPCqAE0bN3
ZaljVNGkPNxFyYgaIwrejyID6rk
-> unsupported

-> done

-> add-identity AGE-PLUGIN-SSS-1R79SSQQQQQQQQQ8L4FTV5NPF26

-> grease-236eb6a331bae717

-> recipient-stanza 0 X25519 DsW
xxx
-> done

-> msg
SGFuZGxlZCBpZGVudGl0eSBpbiAxNzE0MQ
age: sss plugin: Handled identity in 17141
-> ok

-> done

age: error: no identity matched any of the recipients
```

The debug messages I added show that each identity is handled in a different process/pid (each also gets the same stanza separately), so I think the recipients/identities arrays might currently not get longer than 1. It works, but sorting identities won't be possible like this. This is basically https://github.com/FiloSottile/age/issues/526 , which I closed because it seemed to be intentional at the time. I think having only one plugin process with all the data could open more opportunities (more advance logic, ordering, maybe batching encrypts/decrypts in one request/operation, caching pins).

AnomalRoil:
After further testing in `gopass`, I can confirm that for a library trying to support Plugin Recipients and Identities, the current interfaces are not enough yet:
- I had to copy the `parseIdentity` and `parseIdentities` functions to support plugin identities from [cmd/age/parse.go](cmd/age/parse.go)
- I had to copy the `func identitiesToRecipients(ids []age.Identity) ([]age.Recipient, error)` from [cmd/age/age.go](cmd/age/age.go)
- I had to create my own `wrappedIdentity` in order to have a way to output the `encoding` of an identity, it might be best to provide a way of serializing identities easily? (Maybe implement the `Stringer` interface for both `plugin.Identity` and `plugin.Recipient`?)
- While `plugin.Identity` has a `func (i *Identity) Recipient() *Recipient`, it sadly can easily contains secret key material, since it basically contains the plugin.Identity, it would be nice to have a mapping from `plugin.Identity` to `plugin.Recipient` that is "safe", but this probably requires changing the plugin spec and interface to add a new type of interaction, so might not be ideal.

On the plus side:
- implementing my own terminalUI was very easy with the current `ClientUI`, my previous comment about passing it contextes still holds since I had to put `context.Background()` everywhere in mine for now.
- just passing the plugin identities to `age.Decrypt` does work nicely.

One last comment for now, it might be good in [age.Decrypt](https://github.com/FiloSottile/age/blob/176e245b3cb3ada322c21eef0bce166dc5a5e4c7/age.go#L224), when ranging over identities to first sort them to prioritize native identities over plugin ones?

Enzime:
I'm not sure if it is relevant to the scope of this PR, but one thing I think would be useful would be plugin support for `age-keygen`, primarily `age-keygen -y`, this would simplify the CLIs for plugins and make it more consistent

AnomalRoil:
@FiloSottile Any chance of getting this merged soon? 

Or have the security fix related to plugin in it? We're using it in Gopass master atm and are planning a release soon ^^'

I think all of the "bumps on the road" I had using this framework can be solved by additive  improvements that could come later on, such as:
- exporting `parseIdentity` and `parseIdentities`
- exporting `identitiesToRecipients`
- adding `Context` to the UI

PR #594, opened 2024-10-21
build.yml: pattern needs to be '**'

pheiduck:
See: https://github.com/actions/toolkit/blob/main/packages/artifact/docs/faq.md#supported-characters

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Download workflow artifacts
         uses: actions/download-artifact@v4
         with:
-          pattern: age-binaries-*
+          pattern: '**'
           merge-multiple: true
       - name: Upload release artifacts
         run: gh release upload "$GITHUB_REF_NAME" age-*


0x2b3bfa0:
@pheiduck, what was wrong with the previous `age-binaries-*` pattern; didn't it work correctly?

0x2b3bfa0:
The issue you mention ([Foxboron/sbctl@53e074d → .github/workflows/build.yml:60](https://github.com/Foxboron/sbctl/blob/53e074d6934f5ecfffa81a576293219c717f7d19/.github/workflows/build.yml#L60C32-L60C71)) happened because that workflow tries to reference the [`matrix` context](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#matrix-context) in a job where it's not set. If they used a wildcard like we do in `age` it wouldn't have failed:

```diff
-          name: sbctl-binaries-${{ matrix.GOOS }}-${{ matrix.GOARCH }}
+          name: sbctl-binaries-*
```

0x2b3bfa0 (commented):
So, the `age` workflow isn't broken, and this change is functionally superfluous.


<sub>I'd say [if it ain't broke, don't fix it](https://en.wiktionary.org/wiki/if_it_ain%27t_broke,_don%27t_fix_it), but there's nothing inherently wrong with this change; let's wait for official reviews.</sub>

PR #596, opened 2024-10-21
Upgrade go.

Tang8330:
Upgrading Go to address CVEs stemming from `stdlib`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: [1.19.x, 1.x]
+        go: [1.22.x, 1.x]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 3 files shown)


alerque:
Do the CVEs *really* affect this app or do they just happen to exist in an unrelated/unused API surface in Go?

Tang8330:
image

Not sure, but given it's just a Go upgrade, I don't see why we wouldn't just address them.

alerque:
At least one of those has already been brought up here: https://github.com/FiloSottile/age/pull/409#issuecomment-1103827329.

The more general point of "why not just bump" is also addressed in places like here: https://github.com/FiloSottile/age/pull/458#issuecomment-1292662263.

Tang8330:
I understand the position of not wanting to bump for the sake of bumping a CVE that may not actually impact age, thus generating busywork.

However, how do you plan to actually tackle triaging to ensure that appropriate CVEs are being addressed and providing a rationale for why particular CVEs that have been flagged are being purposefully ignored?

FiloSottile:
None of those affect packages that are reachable from this project, so there is nothing to triage.

```
$ govulncheck ./...
=== Symbol Results ===

No vulnerabilities found.

Your code is affected by 0 vulnerabilities.
```

See https://filippo.io/maintenance#on-security-scanners. Please use a better scanner.

PR #591, opened 2024-10-02
feat: move tui code to tui package

brianmcgee:
It makes PluginTerminalUI available for sops (and other third parties) to support plugins more easily.

PR #611, opened 2025-01-13
feat: allow plugin users to create a ClientUI

nicdumz:
Greetings,

I would like to suggest an alternative to
https://github.com/FiloSottile/age/pull/591 which will allow unblocking the
consumer https://github.com/getsops/sops/pull/1641. (I'm interested in plugin
support in `sops` so I can try out FIDO2 keys inside of Nix via `sops-nix`).

Instead of exposing all `age` internals into a separate package (which I
presume isn't the best from a maintenance PoV), I move functions which need to
be shared between `plugin` and `main` into two new internal private modules.

This lets me eventually expose `NewClientUI()` for consumers.

- **Move logging functions to internal/logger.**
- **Move terminal related functions to internal/term**
- **Expose a public `plugin.NewClientUI()`**

Kudos to @brianmcgee for the original work.

@FiloSottile I'm happy to iterate on this code, let me know your constraints and I'll work with you.

HTH.


nicdumz:
And now with :100: passing tests, sorry.

nicdumz:
FWIW, I saw that you were working on a `plugin` feature branch, and it still merges easily with `main` after this, see https://github.com/nicdumz/age/commit/c8259addd76e68985a472ec814de76470477bf90 if you're interested.

FiloSottile:
@nicdumz This was definitely the right approach, thank you!

I implemented it slightly differently, without centralizing the printf functions.

Since it's based on your PR, I kept you as co-author on the commit.

nicdumz:
Sounds good, thank you. We'll wait for the next release and that will allow deleting https://github.com/getsops/sops/blob/main/age/tui.go and replacing it with `plugin.NewClientUI()`

PR #619, opened 2025-02-10
refactor: replace manual slice management with bytes.Buffer in Writer

Gofastasf:
- Simplify buffer management with Write method.
- Use Buffer.Next for moving to next chunk.

### DIFF
--- a/internal/stream/stream.go
+++ b/internal/stream/stream.go
@@ -6,6 +6,7 @@
 package stream
 
 import (
+	"bytes"
 	"crypto/cipher"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


Gofastasf:
@AlexanderYastrebov Hey can you look into this PR :)

AlexanderYastrebov:
@Gofastasf Hello, I am not a maintainer.

Gofastasf:
@FiloSottile Hey can you checkout this PR ?

internal/stream/stream.go
```diff
@@ -215,16 +214,19 @@ const (
...
 	}
 
 	if last {
 		setLastChunkFlag(&w.nonce)
 	}
-	buf := w.a.Seal(w.buf[:0], w.nonce[:], w.unwritten, nil)
-	_, err := w.dst.Write(buf)
-	w.unwritten = w.buf[:0]
+
+	chunk := w.unwritten.Next(ChunkSize)
```
dsemi, line 225: This doesn't release the memory held by the chunk in the buffer. I think you need a `w.unwritten.Reset()` after this. Otherwise the Writes to `w.unwritten` will continue to grow the buffer, which will cause the entire stream content to stay in memory until the Writer is GC'd.

PR #626, opened 2025-03-22
cmd/age: prevent outputting non-printable chars to terminal

ramitmittal:
Buffers decrypted output and errors if it contains non-printable characters.
The method to find non-print chars is to check for the 0 byte, [same as what cURL uses](https://github.com/curl/curl/blob/23f6a1abfd047a8a1b330b4d6898b3a3e5769f6a/src/tool_cb_wrt.c#L207).


FiloSottile:
Thank you for the PR! I've implemented this a bit differently.

PR #627, opened 2025-03-30
Update README.md to add winget

arebee:
Adding in the instructions on how to install with winget on Windows, the native package manager.


arebee:
@FiloSottile Update to README.md assuming that the `winget` entry is yours.

PR #641, opened 2025-06-30
Decrypt with passphrase from stdin

### LINKED ISSUE #603: can't pass  password via stdin during decrypt
## Environment

* OS: Linux 6.9.11
* age version: 1.2.0

## PROBLEM
Currently (afaik) it's not possible to pass passphrase to age via stdin, like:
`echo -n "$passphrase" | age -d -o output_file input_file`

Here, age knows that stdin is NOT used for data, because I specified input file explicitly.

This also does not work with:
`some_cli_password_manager | age -d -i - -o output_file input_file`

Rationale:
1. passphrase is eventually read from stdin
2. sometimes you must use passswords (can't use keys) in script & automation
3. most cli tools allows it (cryptsetup etc.)
4. also #364

AlexanderYastrebov:
Read passphrase from stdin when it is not used for input data.

Fixes #603


cmd/age/age.go
```diff
@@ -455,10 +455,20 @@ func decryptNotPass(flags identityFlags, in io.Reader, out io.Writer) {
 }
 
 func decryptPass(in io.Reader, out io.Writer) {
+	passphrase := passphrasePromptForDecryption
+
+	if in != os.Stdin && !term.IsTerminal(int(os.Stdin.Fd())) {
+		passphrase = func() (string, error) {
+			b, err := io.ReadAll(os.Stdin)
+			b = bytes.TrimRight(b, "\n")
```
AlexanderYastrebov, line 463: Trimming newline to make test case work and improve usability when passphrase is coming from a file or via echo.

Issue #644, opened 2025-07-16
Feature Request (library): Encrypt for io.Reader


FiloSottile:
### Discussed in https://github.com/FiloSottile/age/discussions/638

<div type='discussions-op-text'>

<sup>Originally posted by **dsemi** June 20, 2025</sup>
The age library API is currently implemented to allow for decrypting a Reader, or encrypting a Writer. I have a use case where it would be more convenient to be able to encrypt a Reader to allow for logic to happen in between an encrypted stream and its destination. I guess another way to phrase this would be that decrypting is pull-based, and encrypting is push-based, but I'd like to be able to use pull-based encrypting.

Note that you can sort of work around the issue by using Go's io.Pipe, but that requires an extra goroutine and the error-handling along with it that I'd like to avoid if possible.

I'd be happy to contribute a PR for this if there's no technical limitation as to why this wouldn't work. I took a quick look at internal/stream/stream.go and didn't see any obvious issue, but I might be missing something.
I don't personally have a need for it, but for symmetry I think a decrypting writer could be added as well.</div>

gabyx:
Updated the link to the impl. what we had to do for it to work

PR #646, opened 2025-07-28
Add EncryptReader

AlexanderYastrebov:
Add EncryptReader function to encrypt io.Reader (pull-based encryption) and use it in cmd/age to get rid of io.Copy.

Updates #644

### DIFF
--- a/age.go
+++ b/age.go
@@ -113,6 +113,8 @@ type Stanza struct {
 const fileKeySize = 16
 const streamNonceSize = 16
 
+var errNoRecipients = errors.New("no recipients specified")
+
 // Encrypt encrypts a file to one or more recipients.
 //
 // Writes to the returned WriteCloser are encrypted and written to dst as an age
@@ -122,7 +124,7 @@ const streamNonceSize = 16
 // be encrypted and flushed to dst.
 func Encrypt(dst io.Writer, recipients ...Recipient) (io.WriteCloser, error) {
 	if len(recipients) == 0 {
-		return nil, errors.New("no recipients specified")
+		return nil, errNoRecipients
 	}
 
 	fileKey := make([]byte, fileKeySize)
@@ -130,29 +132,7 @@ func Encrypt(dst io.Writer, recipients ...Recipient) (io.WriteCloser, error) {
 		return nil, err
 	}
 
-	hdr := &format.Header{}
-	var labels []string
-	for i, r := range recipients {
-		stanzas, l, err := wrapWithLabels(r, fileKey)
-		if err != nil {
-			return nil, fmt.Errorf("failed to wrap key for recipient #%d: %v", i, err)
-		}
-		sort.Strings(l)
-		if i == 0 {
-			labels = l
-		} else if !slicesEqual(labels, l) {
-			return nil, fmt.Errorf("incompatible recipients")
-		}
-		for _, s := range stanzas {
-			hdr.Recipients = append(hdr.Recipients, (*format.Stanza)(s))
-		}
-	}
-	if mac, err := headerMAC(fileKey, hdr); err != nil {
-		return nil, fmt.Errorf("failed to compute header MAC: %v", err)
-	} else {
-		hdr.MAC = mac
-	}
-	if err := hdr.Marshal(dst); err != nil {
+	if err := writeHeader(dst, recipients, fileKey); err != nil {
 		return nil, fmt.Errorf("failed to write header: %v", err)
 	}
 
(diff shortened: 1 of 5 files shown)


age.go
```diff
@@ -122,49 +124,81 @@ const streamNonceSize = 16
...
+		return nil, fmt.Errorf("failed to write nonce: %v", err)
+	}
+
+	return stream.NewWriter(streamKey(fileKey, nonce), dst)
+}
+
+// EncryptReader encrypts a reader to one or more recipients.
+//
+// It encrypts src into dst as an age file. Every recipient will be able to decrypt the file.
+func EncryptReader(src io.Reader, dst io.Writer, recipients ...Recipient) error {
```
AlexanderYastrebov, line 153: Should it use (dst, src) order to mimic io.Copy?

AlexanderYastrebov: Return written and/or read size?

AlexanderYastrebov, line 153: TODO: add test

age.go
```diff
@@ -122,49 +124,81 @@ const streamNonceSize = 16
...
 	if _, err := rand.Read(fileKey); err != nil {
 		return nil, err
 	}
 
+	if err := writeHeader(dst, recipients, fileKey); err != nil {
+		return nil, fmt.Errorf("failed to write header: %v", err)
+	}
+
+	nonce := make([]byte, streamNonceSize)
+	if _, err := rand.Read(nonce); err != nil {
```
AlexanderYastrebov, line 140: note: rand.Read never returns error so this and other places could be simplified

internal/stream/stream.go
```diff
@@ -215,16 +215,62 @@ const (
...
+		} else if err == io.ErrUnexpectedEOF {
+			if hasUnwritten {
+				err := writeChunk(dst, aead, &nonce, bufs[unwritten][:ChunkSize], notLastChunk)
+				if err != nil {
+					return err
+				}
+			}
+			return writeChunk(dst, aead, &nonce, bufs[current][:n], lastChunk)
+		} else if err != nil {
+			return err
```
AlexanderYastrebov, line 264: note: here it can do best-effort and write unwritten chunk if any

internal/stream/stream.go
```diff
@@ -215,16 +215,62 @@ const (
...
-	_, err := w.dst.Write(buf)
-	w.unwritten = w.buf[:0]
-	incNonce(&w.nonce)
+	buf := aead.Seal(plaintext[:0], nonce[:], plaintext, nil)
+	_, err := dst.Write(buf)
+	incNonce(nonce)
 	return err
 }
+
+func Encrypt(key []byte, src io.Reader, dst io.Writer) error {
```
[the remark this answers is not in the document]
AlexanderYastrebov: It makes no sense for the stream and user may count written bytes via writer wrapper if needed.

gabyx:
This PR is nice and we need it, would it make sense to add this?

```go
func EncryptStream(src io.Reader, recipients ... Recipient) (dst io.Reader, error) {
	pr, pw := io.Pipe()
	return pr, EncryptReader(pw, src, recipients...)
}
```

We wrap it currently as: https://gitlab.com/data-custodian/custodian/-/blob/88e7389ec5b837672f2b76fd82aaed68a4145e72/components/lib-common/pkg/crypto/reader.go

AlexanderYastrebov:
This PR adds a handy function but what you are looking for is the implementation of https://github.com/FiloSottile/age/issues/644

FiloSottile:
I'm not sure I follow why this API is needed, or why it requires an internals refactor.

Full pull-based encryption like in #644 I understand, and without an `Encrypt(io.Reader) io.Reader` API it requires a goroutine, as you point out.

On the other hand, this `Encrypt(dst io.Writer, in io.Reader)` API only saves a synchronous io.Copy and could be implemented efficiently as

```
func EncryptReader(dst io.Writer, src io.Reader, recipients ...Recipient) error {
	w, err := age.Encrypt(out, recipients...)
	if err != nil {
		return err
	}
	if _, err := io.Copy(w, in); err != nil {
		return err
	}
	return w.Close()
}
```

without any other changes, right? Or am I missing something?

FiloSottile:
Oh I see, it avoids one of the copies inside io.Copy. We could do that also by implementing ReaderFrom / WriterTo, right? If you’re hitting a throughput bottleneck, I can add that to the I/O improvements I’m working on.

AlexanderYastrebov:
> If you’re hitting a throughput bottleneck

TBH, I did not, I just saw a possible optimisation. Making stream.Writer implement ReaderFrom is the better way.

PR #676, opened 2026-01-03
Add darwin amd64 to build workflow

Jimver:
Let's add back darwin amd64 to support Intel Macs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,7 @@ jobs:
           - {GOOS: linux, GOARCH: amd64}
           - {GOOS: linux, GOARCH: arm, GOARM: 6}
           - {GOOS: linux, GOARCH: arm64}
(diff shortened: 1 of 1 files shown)


mynameisbogdan:
> Also, I honestly thought most people used Homebrew on macOS

Some people use mise as it's more easier to have a specific version per repository, without dealing with a single version system wide.

Jimver:
> I don't want to tag a v1.3.2 just for this, because that kicks off a bunch of work for a lot of downstream distributions.

Understandable and that's fine with me! I'll pin it to `1.3.0` for now and will update my project mise configuration on the next release (if this is merged).

yordis:
Thank you for the support, really appreciate.

For what is worth, a lot people discover your package because of `mise` I do think the maintain intent to give you more work but honestly promotes it as a professional solution to use.

I did a one-time donation, hopefully compensates for the frustration, that is the least I can do at the moment.

Thank you so much.

PR #690, opened 2026-02-22
Add PHP implementation reference

xepozz:
Added PHP implementation link to README.


FiloSottile:
Nice! Would you mind adding this to https://github.com/FiloSottile/awesome-age instead?

Also, make sure to integrate the test suite at https://c2sp.org/CCTV/age.

PR #696, opened 2026-03-12
internal/stream: reject sizes below one tag in EncryptedChunkCount

sweis:
`EncryptedChunkCount` returned `(0, nil)` for negative or sub-tag-size inputs —
the consistency check rounds both sides to `0` and passes silently. This would
cascade into `NewDecryptReaderAt` computing `finalChunkIndex = -1`.

Fix: a valid payload always carries at least one Poly1305 tag, even for empty
plaintext. Reject anything smaller than `chacha20poly1305.Overhead` up front.

Test: `go test ./internal/stream/ -run TestEncryptedChunkCountRejectsSmallSizes -v`

### DIFF
--- a/internal/stream/stream.go
+++ b/internal/stream/stream.go
@@ -20,6 +20,9 @@ import (
 const ChunkSize = 64 * 1024
 
 func EncryptedChunkCount(encryptedSize int64) (int64, error) {
+	if encryptedSize < chacha20poly1305.Overhead {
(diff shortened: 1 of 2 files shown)


FiloSottile:
This looks just wrong: I get `invalid encrypted payload size: 0` and `invalid encrypted payload size: 15` from the current code, not `nil`.

PR #697, opened 2026-03-12
cmd/age: avoid int overflow in sshKeyType on 32-bit platforms

sweis:
`sshKeyType` reads a `uint32` length prefix and casts it to `int` for
`cryptobyte.ReadBytes`. On 32-bit platforms a value `≥ 2^31` becomes negative
and cryptobyte panics, so a malformed recipients file can crash the CLI.

Fix: check the length against the remaining buffer before the cast.

Test: `go test ./cmd/age/ -run TestSSHKeyTypeOverflowLength -v`
(uses `recover()` to catch the panic on 32-bit; on 64-bit it asserts graceful
rejection.)

### DIFF
--- a/cmd/age/parse.go
+++ b/cmd/age/parse.go
@@ -124,7 +124,7 @@ func sshKeyType(s string) (string, bool) {
 	k := cryptobyte.String(key)
 	var typeLen uint32
 	var typeBytes []byte
-	if !k.ReadUint32(&typeLen) || !k.ReadBytes(&typeBytes, int(typeLen)) {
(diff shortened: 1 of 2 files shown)


FiloSottile:
> cryptobyte panics, so a malformed recipients file can crash the CLI.

Not true?

```
func (s *String) ReadBytes(out *[]byte, n int) bool {
	v := s.read(n)
	if v == nil {
		return false
	}
```

```
func (s *String) read(n int) []byte {
	if len(*s) < n || n < 0 {
		return nil
	}
```

PR #699, opened 2026-03-12
internal/format: bound header size during Parse

sweis:
`format.Parse` accumulates stanza bodies with no cap on total header bytes, so
a crafted header can exhaust memory in an automated `age.Decrypt` caller before
any MAC is checked.

Fix: wrap the input in a counting reader that errors at 16 MiB. The payload
reader is reconstituted from bufio's overread prepended to the original input
as before; the `rr == input` short-circuit is removed since the wrapper makes
it unreachable.

Test: `go test ./internal/format/ -run TestParseHeaderSizeLimit -v`
(feeds an infinite-stanza-body reader; pre-fix this would read forever.)

### DIFF
--- a/internal/format/format.go
+++ b/internal/format/format.go
@@ -229,11 +229,31 @@ func errorf(format string, a ...any) error {
 	return &ParseError{fmt.Errorf(format, a...)}
 }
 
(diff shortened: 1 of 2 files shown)


FiloSottile:
Opened #701 to discuss this. Feel free to close the PR for now.

PR #707, opened 2026-04-24
feat: x25519 tag and mlkem768x25519 tagpq support

dlubawy:
This PR would add a new x25519 tag and mlkem768x25519 tagpq recipient to `age-plugin-tag` and `age-plugin-tagpq`. I felt the original tagged recipient spec should have included x25519 based keys as well so that users could convert their original age keys to any type of hardware key they desired. I've also been working on [a branch of the age-plugin-yubikey repo](https://github.com/str4d/age-plugin-yubikey/pull/226) which adds support for these key types too, and have been using this code here as a test for that. Having an officially supported reference spec would be great so that I actually know my code there is aligning well with age as a whole.

Not sure if this is desirable since it does add more complication to age which is supposed to be very simple, but I figure it's worth a discussion especially since it's mostly extending the already existing age keys.

### DIFF
--- a/tag/tag.go
+++ b/tag/tag.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Package tag implements tagged P-256 or hybrid P-256 + ML-KEM-768 recipients,
+// Package tag implements tagged P-256/X25519 or hybrid P-256/X25519 + ML-KEM-768 recipients,
 // which can be used with identities stored on hardware keys, usually supported
 // by dedicated plugins.
 //
@@ -28,7 +28,7 @@ import (
 	"golang.org/x/crypto/curve25519"
 )
 
-// Recipient is a tagged P-256 or hybrid P-256 + ML-KEM-768 recipient.
+// Recipient is a tagged P-256/X25519 or hybrid P-256/X25519 + ML-KEM-768 recipient.
 //
 // The latter recipient is safe against future cryptographically-relevant
 // quantum computers, and can only be used along with other post-quantum
@@ -67,7 +67,7 @@ func ParseRecipient(s string) (*Recipient, error) {
 const compressedPointSize = 1 + 32
 const uncompressedPointSize = 1 + 32 + 32
 
-// NewClassicRecipient returns a new P-256 [Recipient] from a raw public key.
+// NewClassicRecipient returns a new P-256/X25519 [Recipient] from a raw public key.
 func NewClassicRecipient(publicKey []byte) (*Recipient, error) {
 	var k hpke.PublicKey
 	var err error
@@ -136,7 +136,7 @@ func (r *Recipient) Tag(enc []byte) ([]byte, error) {
 			}
 		case hpke.MLKEM768X25519().ID():
 			label = "age-encryption.org/mlkem768x25519tag"
-			// In hybrid mode, the tag is computed over just the P-256 part.
+			// In hybrid mode, the tag is computed over just the X25519 part.
 			tagRecipient = tagRecipient[mlkem.EncapsulationKeySize768:]
 			if len(enc) != mlkem.CiphertextSize768+curve25519.PointSize {
 				return nil, fmt.Errorf("invalid ciphertext size")
@@ -161,7 +161,7 @@ func (r *Recipient) Tag(enc []byte) ([]byte, error) {
 }
 
 // WrapWithLabels implements [age.RecipientWithLabels], returning a single
-// "postquantum" label if r is a hybrid P-256 + ML-KEM-768 recipient. This
+// "postquantum" label if r is a hybrid P-256/X25519 + ML-KEM-768 recipient. This
 // ensures a hybrid Recipient can't be mixed with other recipients that would
 // defeat its post-quantum security.
 //
--- a/tag/tag.go
+++ b/tag/tag.go
@@ -124,7 +124,7 @@ func (r *Recipient) Wrap(fileKey []byte) ([]*age.Stanza, error) {
 // identities compatible with tagged recipients.
 func (r *Recipient) Tag(enc []byte) ([]byte, error) {
 	var label string
-	var tagRecipient []byte
+	tagRecipient := r.Bytes()
 	if r.Hybrid() {
 		switch r.pk.KEM().ID() {
 		case hpke.MLKEM768P256().ID():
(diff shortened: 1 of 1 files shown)


str4d:
Replying here instead of on the age-plugin-yubikey PR because it has wider relevance.

> I felt the original tagged recipient spec should have included x25519 based keys as well so that users could convert their original age keys to any type of hardware key they desired.

Key reuse is generally not a desirable property for cryptographic key management. It's also something I am actively designing against while implementing the new native tag and tagpq identity kinds in age-plugin-yubikey, so I'll use them as examples.

The new native tag recipient kind is very similar to the p256tag recipient I defined originally for age-plugin-yubikey (because I helped design it). They both use a P256 key as their core encryption secret, and both include tags in the recipient creating linkability between ciphertexts, so the security properties an age-plugin-yubikey user opted into originally are preserved when migrating to the native tag recipient. Thus it's "fine" for me to auto-upgrade age-plugin-yubikey to change the recipient format without changing the identity in any way. There is still nuance here in that the cryptographic protocol the key is being used for is changing, but with careful attention to domain separation we avoid cross-protocol attacks.

The new native tagpq recipient, by contrast,  requires both a P256 and an ML-KEM 768 secret, and it has a new security property ("be PQ-secure"). Thus I am ensuring that existing age-plugin-yubikey identities can *never* be used as, or upgraded to [0], a tagpq identity. The reason for this is downgrade attacks: if a user can upgrade a legacy identity to tagpq seamlessly, then they do not get PQ security because the identity can still decrypt ciphertexts encrypted to the legacy non-PQ recipient (and retaining that capability would be the point of auto-upgrading). That is unacceptable in a tool for non-cryptographers.

Therefore, I disagree with the suggested additions to age-plugin-yubikey:

- Allowing a legacy native identity to be converted to a "tagged x25519" introduces a ciphertext linkability issue that the original identity creator did not opt into.
- Allowing a legacy native identity to be converted into a "tagged Xwing" introduces a downgrade attack, and thus does not solve the Store-Now-Decrypt-Later attack.
- Allowing a legacy native identity that was generated outside a HW token to be imported into the HW token (which would be necessary as well given I've not yet supported generating legacy native identities there) gives a false sense of security; the key was already exposed outside the HW token, and importing it does not improve security [1].

[0] I cannot technically prevent motivated people from doing this at the moment because of how the ML-KEM secret needs to be stored, but I will not support it in age-plugin-yubikey, I make doing so out-of-band as annoying as possible, and if you do then you lose the legacy or native tag decryption support so it's still not a smooth auto-upgrade.

[1] Yes there are very specific situations where someone might still want to do this for e.g. a particular corporate security posture, and I cannot technically stop them, but I won't support it in age-plugin-yubikey itself because it's not something a general userbase can be expected to do correctly, and thus not something I will accept in my public supported API.

dlubawy:
> Key reuse is generally not a desirable property for cryptographic key management.

I should probably clarify my intent around "convert their original age keys to any type of hardware key they desired". I'm not really trying to encourage key reuse, but rather secure key storage and enable flexible generation/storage capabilities. I guess the first part we might have a fundamental disagreement on based on your third bullet point, but I see converting a key to a hardware key as still being better than just storing that same key as a plaintext file. Yes, the key was exposed outside the secure hardware environment, but the threat actor there is someone with advanced storage recovery capabilities and is more difficult than a text file likely indexed by a file manager (of course, if you can, you should just generate a new key but I'll address this later). The second point isn't really about converting an existing key and more about how someone might want to manipulate key generation for their own needs. I could see there being a reason someone may want bring their own random number generator by using a modification of the reference implementation for age. They then might want to store this on a hardware key as well. This would allow such flexibility with the need to only support one curve type.

Beyond that, I understand your concerns but I think there is some misunderstanding.

>Allowing a legacy native identity to be converted to a "tagged x25519" introduces a ciphertext linkability issue that the original identity creator did not opt into.

My changes don't fully enable this. The changes in this repo makes it possible, but it doesn't make it happen for a user. The same goes for my branch of `age-plugin-yubikey`. It makes it possible, but a user would still need to go through steps outside the project scope to do the conversion themselves. I personally think that's fine because there are sane defaults applied while still giving users the freedom to use their keys how they need. I guess we may disagree on what sorts of freedoms a user should have over their keys.

>Allowing a legacy native identity to be converted into a "tagged Xwing" introduces a downgrade attack, and thus does not solve the Store-Now-Decrypt-Later attack.

My changes in the plugin don't explicitly allow this as I understand this would be a problem too. With that said, my changes are as secure as your own implementation right now. Someone could just upload a Kem seed to the YubiKey data tag slot and then they've "converted" a key. Same way someone could just sign a new cert with an extension containing a KEM seed. I don't think any is really better in that respect, and I would love to hear if there is a better approach to generation/storage on existing hardware.

With that said, my intent right now was to make transfer to a hardware key possible. This is specifically because YubiKeys don't natively support these keys so it has to be generated outside hardware (as far as I'm aware) which is what I alluded to earlier as a potential desire for users (ex: not trusting new PQ libs and generating their keys themselves if they want). This is where that flexibility may be desirable as it allows for creative development of the spec to take place while hardware catches up.

>Allowing a legacy native identity that was generated outside a HW token to be imported into the HW token (which would be necessary as well given I've not yet supported generating legacy native identities there) gives a false sense of security; the key was already exposed outside the HW token, and importing it does not improve security [1].

Again, my changes don't make this happen. It makes it possible, but I don't think that's a problem for the general user base you are concerned about.

I'm not a security expert, so my opinions are probably worthless on those matters. I did at least think a unified spec and having simplicity/flexibility of key usage would be a feature. Sorry to add noise to your project if that's not the case.

Warfront1:
New to these PQ features and trying to get my footing…

It seems to boil down to whether the protocol should restrict X25519 support to prevent misuse, or whether those guardrails belong at the plugin level.

- P-256 and X25519 keys aren't "born" tagged/linkable — they both have the same issue at the protocol level.
- A previously non-linked/tagged P-256 imported key would have the same downgrade surface area.

One group of users not discussed enough is new users interested in using age + Yubikey for PQ purposes.  
There are huge benefits for such users, and the flexibility of this is nice.  
Yubico's docs indicate YubiKeys 5.7+ support on-device X25519 key generation via PIV (I haven't tested this end-to-end).  
A fresh X25519 key generated this way for x25519tag shouldn't have retroactive linkability, downgrade, or import issues — the same clean path P-256 enjoys.  
A fresh mlkem768x25519tag recipient has the same clean security properties — no prior X25519 identity to downgrade to, no import exposure.

The flexibility here would appear to be a feature — this is complicated enough that casual users won't accidentally bypass guardrails, while still being accessible for users who intentionally want manual/direct key management.

PR #713, opened 2026-05-24
Add syntax highlighting to markdown code blocks

codesoap:
Currently, code blocks in README.md and SIGSUM.md don't use any syntax highlighting. I think it does improve readability. Example before:

image

Example after:

image

Unfortunately, the GitHub syntax highlighter doesn't seem perfect and makes mistakes like this:
image

I still think readability is slightly improved overall.


FiloSottile:
Thank you, dropped the couple that were misbehaving.

PR #741, opened 2026-08-30
chore: go fix

zdiff:
`go fix` modernization


FiloSottile:
Will get to it next time that code is touched.

